PR #6, opened 2016-04-19
BACK-1325: Handle authentication errors when deploying.

vseventer:
This PR fixes a bug in the deploy functionality where deploying would fail if the user was not authenticated properly. Due to the project package being pipe-d into the request, the default retry functionality did not work properly. The changes here manually handle authentication errors, so retrying a deploy actually works.


mjsalinger:
LGTM

PR #9, opened 2016-09-19
BACK-1534: Cache the last deploy job ID for ease of retrieval (plus test)

sjl2024:


lib/datalink.coffee
```diff
@@ -71,7 +71,10 @@ class Datalink
     }, (err, response) ->
       if err? then req.emit 'error', err # Trigger request error.
       else # OK.
-        logger.info 'Deploy initiated, received job %s', chalk.cyan response.body.job # Debug.
+        jobResultId = response.body.job
```
vishwakrishnamurthy: you probably don't need this variable at all.
`project.lastJobId = response.body.job` should suffice, no?

lib/datalink.coffee
```diff
@@ -145,6 +148,10 @@ class Datalink
 
   # Returns the deploy job status.
   jobStatus: (job, cb) =>
+    if not job?
+      lastJob = project.lastJobId
+      if not lastJob? then return cb new Error 'No previous job stored. Please provide a job ID.'
+      job = lastJob
```
vishwakrishnamurthy: this can probably be simplified as
```
if not job?
  job = project.lastJobId
  if not job? then return cb new Error 'No previous job stored. Please provide a job ID.'
```

test/job.coffee
```diff
@@ -66,4 +67,10 @@ describe "./#{pkg.name} job", () ->
...
       expect(datalink.jobStatus).to.be.calledOnce
       expect(datalink.jobStatus).to.be.calledWith jobId
-      cb err
\ No newline at end of file
+      cb err
+
+  it 'should print the current job status when called without an id.', (cb) ->
+    job null, command, (err) ->
+      expect(datalink.jobStatus).to.be.calledOnce
+      expect(datalink.jobStatus).to.be.calledWith null
```
vishwakrishnamurthy: shouldn't this be 
`expect(datalink.jobStatus).to.be.calledWith project.lastJobId` ?

sjl2024: No. `datalink.jobStatus` expects the input to be null, in which case it attempts to check for a cached ID

vishwakrishnamurthy: so, how is the cache being tested?

sjl2024: missing tests added in `datalink.test.coffee`

vishwakrishnamurthy:
LGTM. You probably need one more test to ensure the cache is being replaced/updated.

PR #11, opened 2016-09-27
BACK-1836: Modify logs command to display log threshold (if it exists)

sjl2024:
Also added `logs` tests.

### DIFF
--- a/lib/datalink.coffee
+++ b/lib/datalink.coffee
@@ -132,10 +132,13 @@ class Datalink
       if err? then cb err # Continue with error.
       else # Log info.
         logs.forEach (log) ->
-          console.log '%s %s - %s', chalk.green(log.containerId), log.timestamp, chalk.cyan(log.message.trim())
-        logger.info 'Query returned %s logs for Kinvey DLC %s (%s):', chalk.cyan(logs.length), chalk.cyan(project.datalink),
+          if log.threshold?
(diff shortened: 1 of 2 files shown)


lib/datalink.coffee
```diff
@@ -132,10 +132,13 @@ class Datalink
...
-          console.log '%s %s - %s', chalk.green(log.containerId), log.timestamp, chalk.cyan(log.message.trim())
-        logger.info 'Query returned %s logs for Kinvey DLC %s (%s):', chalk.cyan(logs.length), chalk.cyan(project.datalink),
+          if log.threshold?
+            console.log '[%s] %s %s - %s', log.threshold, chalk.green(log.containerId.substring(0, 12)), log.timestamp, chalk.cyan(log.message.trim())
+          else
+            console.log '%s %s - %s', chalk.green(log.containerId.substring(0, 12)), log.timestamp, chalk.cyan(log.message.trim())
+        console.log 'Query returned %s logs for Kinvey DLC %s (%s)', chalk.cyan(logs.length), chalk.cyan(project.datalink),
           chalk.gray(project.datalinkName)
-        cb() # Continue.
+        cb null, logs # Continue.
```
sjl2024: The original behavior is unmodified. The callback was changed to expose log results so I could easily test that the command retrieves the correct data from MAPI. The DLC-CLI doesn't consume the returned data - only the tests do.

PR #24, opened 2016-11-16
BACK-2060: Fix CLI 'ProjectNotConfigured' bug

sjl2024:
Addressed a bug where 'ProjectNotConfigured' was not displayed when attempting to run a command against the default instance while unauthenticated.

### DIFF
--- a/lib/user.coffee
+++ b/lib/user.coffee
@@ -74,9 +74,9 @@ class User
       if data?.host?
         # If `this.host?` at this stage then user has manually updated the host value via the `config` command.
         # Use the new value and neglect the old.
-        unless this.host? then this.host = data.host
+        this.host ?= data.host ? config.host
       else
-        unless this.host? then this.host = config.host
+        this.host ?= data.host ? config.host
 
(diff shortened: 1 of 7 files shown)


sjl2024:
Also contains a directive to check for package updates every 30 minutes instead of every day. This check is done asynchronously and does not impact any CLI command runtime.

lib/user.coffee
```diff
@@ -71,11 +71,12 @@ class User
   restore: (cb) =>
     logger.debug 'Restoring session from file %s', chalk.cyan this.userPath
     util.readJSON this.userPath, (err, data) =>
-      # If we're reading in config from file
       if data?.host?
         # If `this.host?` at this stage then user has manually updated the host value via the `config` command.
         # Use the new value and neglect the old.
         unless this.host? then this.host = data.host
+      else
+        unless this.host? then this.host = config.host
```
mjsalinger: More succinctly written

```
if data?.host?
  this.host ?= data.host ? config.host
```

sjl2024:
@mjsalinger updates pushed

PR #26, opened 2016-12-16
BACK-1883: Adjust client to work with modified server criteria for or…

sjl2024:
…g-backed datalinks (plus removal of deprecated params)


sjl2024:
Update pushed @mjsalinger

PR #29, opened 2017-01-11
BACK-2190: Add protection for parsing/displaying malformed FSR log messages

sjl2024:
This PR adds two features surrounding logging:

- Stringify log messages that are not previously stringified server-side
- When processing and printing server-side log results, skip messages (and add trace code for this when CLI is run in verbose mode) when `message` property is missing

### DIFF
--- a/lib/service.coffee
+++ b/lib/service.coffee
@@ -120,12 +120,17 @@ class Service
     this._execDatalinkLogs from, to, (err, logs) ->
       if err? then cb err # Continue with error.
       else # Log info.
+        skippedLogEntries = 0
         logs.forEach (log) ->
-          if log.threshold?
-            console.log '[%s] %s %s - %s', log.threshold, chalk.green(log.containerId.substring(0, 12)), log.timestamp, chalk.cyan(log.message.trim())
+          if log?.message?
+            if log.threshold?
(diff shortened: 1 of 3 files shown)


lib/service.coffee
```diff
@@ -120,13 +120,18 @@ class Service
     this._execDatalinkLogs from, to, (err, logs) ->
       if err? then cb err # Continue with error.
       else # Log info.
+        skippedLogEntries = []
         logs.forEach (log) ->
-          if log.threshold?
-            console.log '[%s] %s %s - %s', log.threshold, chalk.green(log.containerId.substring(0, 12)), log.timestamp, chalk.cyan(log.message.trim())
+          if log?.message?
```
mjsalinger: Nested if statements can be hard to follow.  Try to avoid whenever you can.  One way would be:

```
unless log.message?
  log.skipped = true
  return skippedLogEntries.push log

if log.threshold?
   rest of for-each function goes here   
```

lib/service.coffee
```diff
@@ -120,13 +120,18 @@ class Service
     this._execDatalinkLogs from, to, (err, logs) ->
       if err? then cb err # Continue with error.
       else # Log info.
```
mjsalinger: `if erR? then return cb err` and remove the `else`.

PR #32, opened 2017-02-17
BACK-1936: es6 rewrite

sjl2024:


mjsalinger:
@sjl2024 Can you address the conflicts?

lib/project.js
```diff
@@ -0,0 +1,228 @@
...
+    this.serviceName = null;
+    this.schemaVersion = null;
+    this.lastJobId = null;
+
+    this._selectService = (this._selectService).bind(this);
+    this._selectOrg = (this._selectOrg).bind(this);
+    this._selectAppOrOrg = (this._selectAppOrOrg).bind(this);
+    this._selectApp = (this._selectApp).bind(this);
+    this._execKinveyServices = (this._execKinveyServices).bind(this);
+    this.setup = (this.setup).bind(this);
```
mjsalinger: Are these binds necessary?  Since `setup`, `select`, etc are already class members, aren't they implicitly bound to the `this` of the class?

lib/service.js
```diff
@@ -0,0 +1,257 @@
...
+const STATUS_CONSTANTS = {
+  ONLINE: 'ONLINE',
+  NEW: 'NEW',
+  UPDATING: 'UPDATING',
+  ERROR: 'ERROR'
+};
+
+class Service {
+  constructor() {
+    this.serviceStatus = (this.serviceStatus).bind(this);
```
mjsalinger: Again, what's the point of the bind?  Seems like unnecessary overhead...

lib/user.js
```diff
@@ -0,0 +1,165 @@
...
+const request = require('./request.js');
+const util = require('./util.js');
+
+class User {
+  constructor(path) {
+    this.token = null;
+    this.tokens = {};
+    this.host = null;
+
+    this.setup = (this.setup).bind(this);
```
mjsalinger: Same bind question as above.

lib/project.js
sjl2024: They weren't implicitly bound. I'll look into this a bit more (it was awhile ago), but simply using es6 classes wasn't directly compatible based on the coffee mechanisms used to originally implement `project.js` and `user.js`

sjl2024:
@mjsalinger we can't resolve the conflicts because it's a rewrite. Those are coffee changes.. I've maintained a list of tickets added to master while this rewrite marinated. I'll ensure to port them into this branch before we force merge it into master.

andrasq:
the methods are unbound from the `project` object in `select()`,
where the bare function bodies are passed to `async.series`.
For this construct to work, the methods should be wrapped in
functions, eg `(next) => this._selectAppOrg(next)`

sjl2024:
@mjsalinger Un-hung the deploy tests, addressed @andrasq's feedback (removed all `bind` calls), and made some formatting changes. Ready for another look

PR #36, opened 2017-07-20
Back 1702 provide credentials externally

DennyGD:


lib/constants.js
```diff
@@ -0,0 +1,23 @@
...
+ * This software contains valuable confidential and proprietary information of
+ * KINVEY, INC and is subject to applicable licensing agreements.
+ * Unauthorized reproduction, transmission or distribution of this file and its
+ * contents is a violation of applicable laws.
+ */
+
+const constants = {
+  EnvironmentVariables: {
+    User: 'KINVEY_USER',
+    Password: 'KINVEY_PASSWORD'
```
sjl2024: We have some other Kinvey command line utilities, and while I doubt the values used here would potentially pose a conflict, it would be a good idea to be slightly more verbose. E.g. `KINVEY_CLI_USER`, `KINVEY_CLI_PASSWORD`

Edit: Also, this format is consistent with the CLI session file name: https://github.com/Kinvey/kinvey-cli/blob/master/config/default.js#L28

lib/user.js
```diff
@@ -39,11 +40,23 @@ class User {
     return this.token;
   }
 
-  login(email, password, cb) {
+  login(email, password, done) {
+    const envEmail = process.env[constants.EnvironmentVariables.User];
+    const envPassword = process.env[constants.EnvironmentVariables.Password];
+    const useEnvVariables = !email && !password && envEmail && envPassword;
```
sjl2024: probably don't need to create a variable for this test since it's only used in a single place

test/mocha.opts
```diff
@@ -1,5 +1,4 @@
 --check-leaks
---compilers coffee:coffee-script/register
```
sjl2024: Nice catch!

test/user.js
```diff
@@ -18,6 +18,28 @@ const api = require('./lib/api.js');
 const prompt = require('../lib/prompt.js');
 const user = require('../lib/user.js');
 const util = require('../lib/util.js');
+const constants = require('./../lib/constants');
+
+const invalidEmail = 'invalid@example.com';
```
sjl2024: Test constants/values should be moved to the appropriate `describe` block and initialized in a `before/beforeEach` block on the `this` namespace. An example:

```
describe('user', () => {
  beforeEach('credential setup', () => {
    this.invalidEmail = 'invalid@example.com';
    ...
  });
});
```

test/user.js
```diff
@@ -323,11 +380,15 @@ describe('user', () => {
     });
   });
   describe('setup', () => {
+    const email = 'bob@example.com';
```
sjl2024: These values should be moved into a `before/beforeEach` block

test/user.js
```diff
@@ -18,6 +18,28 @@ const api = require('./lib/api.js');
 const prompt = require('../lib/prompt.js');
 const user = require('../lib/user.js');
 const util = require('../lib/util.js');
+const constants = require('./../lib/constants');
+
+const invalidEmail = 'invalid@example.com';
+const invalidPassword = 'invalidPass';
+
+const helper = {
```
sjl2024: I'm not sure that these functions need to be nested inside of an object (since it's not being exported anywhere) but if you choose to leave it this way, I find that it's best to be as descriptive as possible with variable naming - e.g. `loginTestsHelper` - especially since this is public-facing code

lib/user.js
```diff
@@ -39,11 +40,23 @@ class User {
     return this.token;
   }
 
-  login(email, password, cb) {
+  login(email, password, done) {
```
DennyGD: Old habits die hard I guess :smile:
Fixed

lib/user.js
DennyGD: You're right. I just find the intent of the if statement more clear this way

test/user.js
DennyGD: The 'invalidEmail' is used throughout the tests and its value never changes. Could you be more specific as to why you think it should be moved and attached to this?

test/user.js
DennyGD: As with the 'invalidEmail' variable, this also doesn't change. Its scope is the 'setup' describe. I don't really mind moving it but I'd like to understand the reasons behind it :confused:

sjl2024:
LGTM. Please document the new feature in the README and rebase and afterward I'll merge.

test/user.js
sjl2024: I'm fine with leaving `const` for test values provided that they're scoped to the nearest-applicable `describe`.

README.md
```diff
@@ -31,6 +31,21 @@ In your project directory, run `kinvey config` to set-up your project. The CLI w
 
 You can create a one-time session using the `-e` and `-p` (and optionally `--host`) flags. Set these parameters explicitly in your calls if you are unable to init with `kinvey config [host]`.
```
sjl2024: This section is actually related to credentials, and is a bit redundant now that you've added a full section on them. I would look at either a) removing it or b) incorporating it into the ### Credentials section as you see fit

README.md
```diff
@@ -31,6 +31,21 @@ In your project directory, run `kinvey config` to set-up your project. The CLI w
...
 
+### Credentials
+
+Credentials can be specified either as command line options or as environment variables.
+* command line options
```
sjl2024: I think sub-sections would be better here (instead of a bullet list) so that you can move the example above 'You can create a one-time session...' down here.

An example: 

#### Command line options
Run `kinvey config` and the CLI...
You can create a one-time-session using the `-e` and `-p` (and optionally `--host`)

Example:
```kinvey config -e kinveyAccount@kinvey.com -p yourKinveyPassword`

#### Environment variables
* `KINVEY_CLI_USER` - e-mail address ...
* `KINVEY_CLI_PASSWORD` password of your...

Example:
```
export KINVEY_CLI_USER=kinveyAccount@example.com && export KINVEY_CLI_PASSWORD=yourKinveyPassword
```

PR #38, opened 2017-08-04
Back 2704 unit tests improvements

DennyGD:
Added tests coverage
Increased coverage for some modules
Added some constants
Tests refactoring


cmd/config.js
```diff
@@ -20,6 +20,7 @@ const logger = require('../lib/logger.js');
 const project = require('../lib/project.js');
 const user = require('../lib/user.js');
 const util = require('../lib/util.js');
+const handleActionFailure = require('./../lib/util').handleActionFailure;
```
sjl2024: Nit: I would rename `handleActionFailure` to `handleCommandFailure` since it's more specific

DennyGD:
Changed tests structure.
Current tests coverage stats:

sjl2024:
@DennyGD LGTM. Looking it over one more time, it looks like some functions in util.js `readFile` are only aren't used anywhere else in the code and could probably be removed. Also, despite good coverage on util's APIs, it's a utility library so it would make sense to create standalone unit tests for it. But with the impending rewrite and integration tests, it might not be necessary - I'll leave it up to you

DennyGD:
I hope to get a chance to refactor util.js during the refactor and add unit tests then.

PR #40, opened 2017-08-24
BACK-2804: Initial refactor of logs command

sjl2024:
- Adds `--start` and `--end` flags (which replace) `kinvey logs [from] [to]` optional method params
- Adds `--number`/`-n` option for specifying number of entries to return (default = 100, max = 2500)
- Adds `--page` option for viewing logs in pages (of size `-n`)

... plus tests

### DIFF
--- a/cmd/logs.js
+++ b/cmd/logs.js
@@ -14,6 +14,7 @@
  */
 
 const async = require('async');
+const config = require('config');
 const moment = require('moment');
 const program = require('commander');
 const service = require('../lib/service.js');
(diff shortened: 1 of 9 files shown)


cmd/logs.js
```diff
@@ -28,14 +29,25 @@ function validateTimestamp(ts) {
   return moment(ts, moment.ISO_8601, true).isValid();
 }
 
-function logs(from, to, command, cb) {
+function validateNumber(number) {
+  if (number == null) return true;
+  if (number === 0 || number === '0') return false;
```
DennyGD: Maybe, func name could be changed to something a bit more descriptive or a comment be added? Like isValidInteger, isValidNonZeroInteger...

cmd/logs.js
```diff
@@ -28,14 +29,25 @@ function validateTimestamp(ts) {
...
+function logs(command, cb) {
   const options = init(command);
-  if (!validateTimestamp(from)) return cb(new Error('Logs \'from\' timestamp invalid (ISO-8601 required)'));
-  if (!validateTimestamp(to)) return cb(new Error('Logs \'to\' timestamp invalid (ISO-8601 required)'));
+
+  // Validate input parameters
+  if (!validateTimestamp(options.start)) return handleActionFailure(new Error('Logs \'start\' timestamp invalid (ISO-8601 expected)'), cb);
+  if (!validateTimestamp(options.end)) return handleActionFailure(new Error('Logs \'end\' timestamp invalid (ISO-8601 expected)'), cb);
+  if (!validateNumber(options.number)) return handleActionFailure(new Error('Logs \'number\' parameter invalid (non-zero integer expected)'), cb);
+  if (!validateNumber(options.page)) return handleActionFailure(new Error('Logs \'page\' parameter invalid (non-zero integer expected)'), cb);
```
DennyGD: If these error messages get extracted as some constants, they could be used in the tests, too, and easily changed

lib/util.js
```diff
@@ -100,7 +100,7 @@ function makeRequest(options, cb) {
         });
       }
     }
-    if (response.body != null && response.body != null) return cb(new KinveyError(response.body.code, response.body.description));
+    if (response.body != null && response.body != null) return cb(new KinveyError(response.body.code, response.body.description, response.body.debug));
```
DennyGD: I know it's not your change but do you think you can remove one of the `response.body` evaluations? :satisfied:

sjl2024:
@DennyGD updates pushed

PR #41, opened 2017-09-14
BACK-2579: Display service version as part of status command (if vers…

sjl2024:
…ion is provided in server response)

### DIFF
--- a/lib/service.js
+++ b/lib/service.js
@@ -160,21 +160,35 @@ class Service {
   serviceStatus(cb) {
     return this._execServiceStatus((err, response) => {
       if (err != null) return cb(err);
+
       let status = response.body.status;
+      let version = response.body.version;
+
+      if (version != null) {
+        version = chalk.whiteBright(`v${version}`);
+      }
+
       if (status === STATUS_CONSTANTS.ONLINE) {
-        status = chalk.green(STATUS_CONSTANTS.ONLINE);
+        status = chalk.greenBright(STATUS_CONSTANTS.ONLINE);
       }
       if (status === STATUS_CONSTANTS.UPDATING) {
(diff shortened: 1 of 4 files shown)


lib/service.js
```diff
@@ -160,21 +160,35 @@ class Service {
...
-      return cb(null, response.body.status);
+
+      // Print version if attached to response
+      if (response.body.version != null) {
+        logger.info('Service status: %s [%s]', status, version);
+      } else {
+        logger.info('Service status: %s', status);
+      }
+
+      return cb(null, response.body.status, response.body.version);
```
DennyGD: Wouldn't it be better if the callback receives only two arguments - error as first and data as second? It seems more like the node.js way. And data would be an object containing status, version and whatever else may be added in the future.

lib/service.js
```diff
@@ -160,21 +160,35 @@ class Service {
...
       if (status === STATUS_CONSTANTS.UPDATING) {
-        status = chalk.yellow(STATUS_CONSTANTS.UPDATING);
+        status = chalk.yellowBright(STATUS_CONSTANTS.UPDATING);
       }
       if (status === STATUS_CONSTANTS.NEW) {
-        status = chalk.cyan(STATUS_CONSTANTS.NEW);
+        status = chalk.cyanBright(STATUS_CONSTANTS.NEW);
       }
       if (status === STATUS_CONSTANTS.ERROR) {
-        status = chalk.red(STATUS_CONSTANTS.ERROR);
```
DennyGD: We should change these if statements into switch case or if-else if :grinning: Haven't noticed it so far but status is either online or updating or whatever. Anyways, we can do it any time.

DennyGD:
LGTM

PR #42, opened 2017-09-14
BACK-2804: Error out if deprecated 'from' and/or 'to' params are provided to 'logs' cmd

sjl2024:
This branch gracefully handles the deprecation of `[from] [to]` parameters (as part of the transition to `--from --to` flags in version 2.0).
* Added `from` and `to` args back to the `logs` command function to prevent the CLI from crashing if legacy users try to filter logs using the deprecated parameters
* CLI throws an error instructing users to use the new command format when either deprecated param is used
* Added tests

### DIFF
--- a/cmd/logs.js
+++ b/cmd/logs.js
@@ -36,7 +36,7 @@ function isValidNonZeroInteger(number) {
   return /^\d+$/.test(number);
 }
 
-function logs(command, cb) {
+function logs(from, to, command, cb) {
   const options = init(command);
 
   // Validate input parameters
@@ -56,7 +56,12 @@ function logs(command, cb) {
   return async.series([
     (next) => user.setup(options, next),
     (next) => project.restore(next),
-    (next) => service.logs(options.start, options.end, options.number, options.page, next)
+    (next) => service.logs(options.start, options.end, options.number, options.page, next),
+    (next) => {
+      if (from != null) logger.warn('Logs \'from\' param is deprecated. Please use the \'--start\' flag instead');
+      if (to != null) logger.warn('Logs \'to\' param is deprecated. Please use the \'--end\' flag instead');
+      next();
+    }
   ], (err) => {
     handleActionFailure(err, cb);
   });
--- a/cmd/logs.js
+++ b/cmd/logs.js
@@ -38,12 +38,12 @@ function isValidNonZeroInteger(number) {
   return /^\d+$/.test(number);
 }
 
-function logs(from, to, command, cb) {
+function logs(argsArray, command, cb) {
   const options = init(command);
 
   // Handle deprecated logs command params
-  if (from != null || to != null) {
-    return handleActionFailure(new KinveyError('DeprecationError', `Logs ${chalk.whiteBright('[from]')} and ${chalk.whiteBright('[to]')} parameters have been removed. Use ${chalk.blueBright('--from')} and ${chalk.blueBright('--to')} flags instead`), cb);
+  if (argsArray != null && argsArray.length > 0) {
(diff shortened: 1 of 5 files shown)


sjl2024:
@mjsalinger updates pushed based on team discussion

sjl2024:
Updates after team discussion in backend-internal:
*`--start` and `--end` flags reverted back to `--to` and `--from` (for platform consistency purposes)
* CLI now errors out (instead of logging warnings) when `kinvey logs` is called with the deprecated parameters (e.g. `kinvey logs [from] [to]`)

cmd/logs.js
```diff
@@ -36,27 +38,29 @@ function isValidNonZeroInteger(number) {
...
-    return handleActionFailure(new Error(`Logs \'start\' ${LogErrorMessages.INVALID_TIMESTAMP}`), cb);
-  }
-  if (!isValidTimestamp(options.end)) {
-    return handleActionFailure(new Error(`Logs \'end\' ${LogErrorMessages.INVALID_TIMESTAMP}`), cb);
-  }
-  if (!isValidNonZeroInteger(options.number)) {
-    return handleActionFailure(new Error(`Logs \'number\' ${LogErrorMessages.INVALID_NONZEROINT}`), cb);
+  // Handle deprecated logs command params
+  if (argsArray != null && argsArray.length > 0) {
+    return handleActionFailure(new KinveyError('DeprecationError', `Version 1.x ${chalk.whiteBright('[from]')} and ${chalk.whiteBright('[to]')} params have been converted to options. Use ${chalk.blueBright('--from')} and ${chalk.blueBright('--to')} to filter by timestamp instead.`), cb);
```
DennyGD: I find it a bit odd that you 'paint' some of the words. It's like you have prior knowledge when you are creating the error that it will be logged. And you shouldn't have because it is up to the `handleActionFailure` function to determine what to do with the error. Also, none of the other `new Error/KinveyError` invocations include colored strings. Is it really necessary here?
If it turns out, it is, then I think we'll have to consider doing some changes during the refactoring. :pensive:

test/cmd/logs.test.js
```diff
@@ -44,59 +44,67 @@ describe(`./${pkg.name} logs`, () => {
...
 
   it('should retrieve log entries based on query', (cb) => {
-    logs(command, (err) => {
+    logs(null, command, (err) => {
       expect(service.logs).to.be.calledOnce;
       cb(err);
     });
   });
 
   it('should fail with an invalid \'start\' timestamp', (done) => {
```
DennyGD: from

test/cmd/logs.test.js
```diff
@@ -44,59 +44,67 @@ describe(`./${pkg.name} logs`, () => {
...
-    logs(command, (err) => {
+    command.addOption('from', 'abc');
+    logs(null, command, (err) => {
       expect(err).to.exist;
       expect(err.message).to.contain(LogErrorMessages.INVALID_TIMESTAMP);
       done();
     });
   });
 
   it('should fail with an invalid \'end\' timestamp', (done) => {
```
DennyGD: to?

cmd/logs.js
sjl2024: I don't think the latter case is far-fetched.

My rationale for using colors here is because of the 'hard-line' deprecation. Typically when deprecating things you give users some 'advance' notice. For example, allow `[to]` and `[from]` params as they'd normally function, but print warnings when they're used. In this case we removed them altogether, without any deprecation warnings, so I wanted to emphasize that these parameters weren't removed, but converted to flags.

My opinion is that in this case the colors are okay, because it's a 'special' case, but more importantly, in a future version the command parameters (including the colorful error message) will be removed altogether (probably when the rewrite occurs). Do you think they're okay given these circumstances? If you still have doubts then I have no issue with removing them either. Just trying to do what's best for users

DennyGD: :joy: It haven't occurred to me that we will remove the message, actually. Fair enough

PR #47, opened 2017-10-04
BACK-2580: Show who deployed latest Flex service and when

samlevin:
Enhances the `status` command to show the deployer email address (and first/last name, if set) and the formatted date/time of the most recent deploy. Plus tests.

### DIFF
--- a/lib/service.js
+++ b/lib/service.js
@@ -173,18 +173,40 @@ class Service {
         status = chalk.redBright(ServiceStatus.ERROR);
       }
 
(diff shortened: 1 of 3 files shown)


samlevin:
Per a discussion with @DennyGD the integration test for the `status` command is skipped in this branch because it knew too much about the command implementation and is no longer practical to use as a testing mechanism. This will be addressed in the CLI refactor (in which the integration tests will also be refactored).

PR #48, opened 2017-10-05
BACK-2892: Replace commander with yargs

DennyGD:
Removed commander.js and started using yargs
Fixed the tests accordingly
Skipped cli.test.js where options such as verbose and silent are checked. Makes more sense to fix these tests after the namespaces and the refactoring.
Tried to restrain myself from doing any other changes


cmd/list.js
```diff
@@ -15,25 +15,23 @@
...
 const init = require('../lib/init.js');
-const program = require('commander');
 const project = require('../lib/project.js');
 const user = require('../lib/user.js');
 const handleActionFailure = require('./../lib/util').handleCommandFailure;
 
-function list(command, cb) {
-  const options = init(command);
+function list(argv, cb) {
+  init(argv);
```
samlevin: Can we handle this `init` behavior centrally, as opposed to in every command? I realize it was like this before but maybe now would be a good time to split it up in order to separate concerns, improve testability, etc.

cmd/logout.js
```diff
@@ -14,12 +14,11 @@
...
 
 const async = require('async');
-const program = require('commander');
 const project = require('../lib/project.js');
 const user = require('../lib/user.js');
 const handleActionFailure = require('./../lib/util').handleCommandFailure;
 
-function logout(command, cb) {
+function logout(argv, cb) {
   return async.series([
```
samlevin: Is `init()` intentionally omitted from this command?

DennyGD: IMO it shouldn't be but it was like this. I intend to change it during the refactoring, though.

cmd/list.js
DennyGD: Yeah, I was thinking the same but I'm not sure we can do it... I mean, we probably can one way or another but I believe `yargs` will introduce middlewares in the future. And that's a thing exactly for a middleware. But we can discuss it again during the refactoring.

DennyGD:
Tested manually. The comments will be addressed probably during the refactoring, so I'll just merge (Sam approved on slack).

PR #51, opened 2017-10-27
Back 2893 profile namespace and init

DennyGD:
General refactoring:
* introduce CLI manager to stitch everything together and to prevent modules from communicating directly with each other
* pass dependencies in constructor or as arguments
* move most of the authentication-related logic to a one central place (check function from yargs)

Prepare to handle the profiles namespace

Logout the user when it is a one-time session

Implement 'kinvey init' to create a profile through prompts

Implement the profile namespace

Support environment variables (BACK-2780)

Add support for profile option

Add support for serviceId option in flex

Sorry for the multiple changes.


dokov:
I looked through the code and I have the following general comments/suggestions:

- I think the current file structure is not optimal. Currently there is a folder "commands" with sub-folders for each command and those contain just the definitions of the commands (name, description, etc). The actual functionality of the command is in under a sub-folder of a different folder "controllers" (again sub-folders for each command). I think it would be better to have only folders for each command and place in it everything that is related to the specific command.

- Currently the definition of a command is spread along files for each sub-command. Those files are always just a few lines as they do not contain functionality, just the name of the sub-command and description. I would prefer having all the sub-commands in a single file - will be easier to manage.

Let me know what you think.

Otherwise code looks good to me. Note, that I am not familiar with the previous state, I only looked over the code as it is now.

@samlevin Can you do a code review as well?

lib/authentication.js
```diff
@@ -0,0 +1,97 @@
...
+
+    this.cliManager.sendRequest(
+      {
+        data,
+        host: host,
+        endpoint: Endpoints.session(),
+        method: HTTPMethod.POST
+      },
+      (err, result) => {
+        debugger;
```
dokov: I believe this should not be committed.

DennyGD:
@dokov  I've tried moving the sub-commands into a single file today but it became a bit messy and I've dropped it. `logs.js` for example is bigger and it contains several options and a builder function. But I don't like having so many files either.  

And if we add more options or more builder functions to the sub-commands, the file will grow. I'm 50/50, so if you find it better this way, I'll change it.

samlevin:
@DennyGD I saw that you made some commits recently. Should I wait to review? cc @dokov

DennyGD:
@samlevin I made some small commits including the one mentioned at standup yesterday (to fix lint errors). After all, there is always room for improvement. No need to wait  :smile:

lib/prompter.js
```diff
@@ -0,0 +1,67 @@
...
+    });
+  }
+
+  /**
+   * Builds a question.
+   * @param message
+   * @param name
+   * @param type
+   * @param defaultValue
+   * @param choices
```
samlevin: Should add param descriptions when using JSDoc

lib/request.js
```diff
@@ -13,15 +13,105 @@
...
-  timeout: config.timeout || 5000
-};
+const config = require('config');
+const { Errors, HTTPMethod } = require('./constants');
+const KinveyError = require('./kinvey-error');
+const { isConnectionError, isNullOrUndefined } = require('./utils');
+
+class Request {
+  constructor(user, options) {
+    debugger;
```
samlevin: remove `debugger` everywhere before release

lib/request.js
```diff
@@ -13,15 +13,105 @@
...
-const options = {
-  baseUrl: config.host,
-  json: true,
-  timeout: config.timeout || 5000
-};
+const config = require('config');
+const { Errors, HTTPMethod } = require('./constants');
+const KinveyError = require('./kinvey-error');
+const { isConnectionError, isNullOrUndefined } = require('./utils');
+
```
samlevin: The CLI is growing quite a bit, I think it could be helpful to start adding general descriptions at the top of each file to describe its purpose (with or without JSDoc - your call)

package.json
```diff
@@ -35,8 +35,9 @@
     "custom-logger": "0.3.1",
     "inquirer": "0.10.0",
     "isemail": "1.2.0",
+    "lodash.isempty": "4.4.0",
```
samlevin: This extra dependency can be omitted by using `return (input == null)` in `utils.isNullOrUndefined`. That's exactly what `isnil` does: https://github.com/lodash/lodash/blob/6cb3460fcefe66cb96e55b82c6febd2153c992cc/isNil.js#L19

package.json
```diff
@@ -35,8 +35,9 @@
...
+    "lodash.isnil": "4.0.0",
```
samlevin: Same comment above could apply to this dependency

lib/utils.js
```diff
@@ -0,0 +1,247 @@
...
+  },
+  serviceStatus: (schemaVersion, serviceId) => {
+    return `${utils.Endpoints.services(schemaVersion, serviceId)}/status`;
+  },
+  serviceLogs: (schemaVersion, serviceId) => {
+    return `${utils.Endpoints.services(schemaVersion, serviceId)}/logs`;
+  }
+  // `/v${project.schemaVersion}/data-links/${project.service}/status`
+
+  // `/v${this.schemaVersion}/${resourceType}/${entity}/data-links`
```
samlevin: Commented out lines should be removed

lib/setup.js
```diff
@@ -0,0 +1,82 @@
...
+    const data = {
+      active: this._activeItems,
+      profiles: this._profiles
+    };
+
+    writeJSON(this._path, data, done);
+  }
+
+  getProfiles() {
+    // not cool
```
samlevin: Comment should be clarified or removed before release

lib/setup.js
```diff
@@ -0,0 +1,82 @@
+const { isEmpty, isNullOrUndefined, readJSONSync, writeJSON } = require('./utils');
+
+// TODO: think about validation
```
samlevin: Is there a ticket for this?

lib/prompter.js
DennyGD: Well, they are optional really...  :worried: Sometimes, the name is self-explanatory and in this case the class is just a wrapper around inquirer.

lib/request.js
DennyGD: Done

lib/project-setup.js
```diff
@@ -0,0 +1,66 @@
...
+const { isEmpty, isNullOrUndefined, readJSONSync, writeJSON } = require('./utils');
+
+class ProjectSetup {
+  constructor(path) {
+    this._path = path;
+    this._flexNamespace = {};
+  }
+
+  save(done) {
+    const data = {
```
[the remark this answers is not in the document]
DennyGD: Changed `path` to `projectPath`.

DennyGD:
@samlevin `npm test` is expected to fail at this moment. There's a separate ticket about the tests. After we merge this into the feature branch, I'll open a PR for the integration tests. Still haven't written unit tests because I wasn't sure what changes would be needed after the code review.

Branch is ready to be reviewed again.

cc @dokov

package.json
DennyGD: Removed `isnil` but `isempty` is still there because it contains a bit more logic.

lib/authentication.js
```diff
@@ -0,0 +1,95 @@
...
+    return this._currentUser && this._currentUser.token;
+  }
+
+  login(email, password, MFAToken, host, done) {
+    const data = {
+      email,
+      password
+    };
+
+    if (MFAToken) {
```
samlevin: `if (thing)` should only be used if `thing` is expected to be `true` or `false` (i.e. `if (thing === true)`). For null/undefined, explicit checks `if (MFAToken != null)` should be used instead. This applies everywhere in the codebase... (e.g. `if (err != null)`

lib/project-setup.js
```diff
@@ -0,0 +1,66 @@
...
+    writeJSON(this._path, data, done);
+  }
+
+  load() {
+    try {
+      const projectSetup = readJSONSync(this._path);
+      if (!isEmpty(projectSetup)) {
+        this._flexNamespace = projectSetup.flex;
+      }
+    } catch (ex) {
```
samlevin: convention is (`err`)

lib/project-setup.js
```diff
@@ -0,0 +1,66 @@
...
+      return ex;
+    }
+  }
+
+  isFlexSet() {
```
samlevin: Which part of Flex does this refer to? Setup/config/service ID/etc. (ambiguous name)

lib/controllers/init-controller.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  // TODO: refactor this nonsense
+  handleInit(done) {
+    let retryCreation = false;
+    let input;
+
+    async.series([
+      (next) => {
+        this._getMandatoryUserInput((err, data) => {
```
samlevin: Noticed that this is called `getMandatoryUserInput` here but `getMandatoryQuestions` elsewhere. Should we choose one or the other for consistency?

lib/controllers/init-controller.js
```diff
@@ -0,0 +1,127 @@
...
+      PromptTypes.INPUT,
+      null,
+      null,
+      utils.validateMFAToken
+    );
+
+    this.cliManager.prompter.prompt(question, done);
+  }
+
+  // TODO: refactor this nonsense
```
samlevin: Should be resolved before release

lib/controllers/flex-controller.js
```diff
@@ -0,0 +1,270 @@
...
+        });
+      }
+    ], (err, results) => {
+      const result = results.pop();
+      this.cliManager.processCommandResult(err, result, done);
+    });
+  }
+
+  deleteProjectSetup(done) {
+    const err = this.ensureFlexSetup();
```
samlevin: If this is synchronous you should consider using `try`/`catch` here (convention/easier to understand).

lib/cli-manager.js
```diff
@@ -0,0 +1,352 @@
...
+    this._authentication.login(email, password, MFAToken, host, done);
+  }
+
+  createProfile(profileName, email, password, MFAToken, host, done) {
+    // TODO: decide what to do if profile already exists
+    if (utils.isNullOrUndefined(profileName)) {
+      return done(new Error('Profile name is not set.'));
+    }
+
+    const self = this;
```
samlevin: Is there a specific reason for using `const self = this` instead of using arrow functions in the `async.series` array?

lib/cli-manager.js
```diff
@@ -0,0 +1,352 @@
...
+
+  /**
+   * Sets the current user if not already set using info from options.
+   * @param {Object} options
+   * @param done
+   * @returns {*}
+   */
+  setCurrentUserFromOptions(options, done) {
+    if (this._authentication.hasCurrentUser()) {
+      return done(null);
```
samlevin: `return done();`

lib/cli-manager.js
```diff
@@ -0,0 +1,352 @@
...
+    });
+  }
+
+  log(logLevel, ...args) {
+    this._logger[logLevel](...args);
+  }
+
+  _clearSession(done) {
+    if(!this._isOneTimeSession) {
+      return done(null);
```
samlevin: `return done()`

lib/cli-manager.js
```diff
@@ -0,0 +1,352 @@
...
+
+        this._loadGlobalSetup();
+
+        if (!this._isOneTimeSession && this._requiresAuth) {
+          this._setCurrentUserFromProfile(argv[AuthOptionsNames.PROFILE]);
+        }
+
+        if (this._requiresAuth && !this._isOneTimeSession && !this._authentication.hasCurrentUser()) {
+          // TODO: change err msg
+          throw new Error('You must be authenticated. (CHANGE ERR MSG)');
```
samlevin: `(CHNGE ERR MSG)` to what? :)

lib/cli-manager.js
```diff
@@ -0,0 +1,352 @@
...
+        }
+      )
+      .option('verbose', {
+        global: true,
+        describe: 'Output debug messages',
+        type: 'boolean'
+      })
+      .check((argv) => {
+        // FIXME: Move to a middleware when yargs provides middleware support (https://github.com/yargs/yargs/pull/881)
+        debugger;
```
samlevin: Debug util should be removed

lib/cli-manager.js
```diff
@@ -0,0 +1,352 @@
...
+    const err = this._setup.load();
+    const canContinue = !err || err instanceof Error && err.code === 'ENOENT';
+    if (!canContinue) {
+      this._logger.error('Failed to load global settings.');
+      throw err;
+    }
+  }
+
+  _initCommandsManager(args) {
+    if (args) {
```
samlevin: Per the style guide (https://github.com/Kinvey/style-guides/blob/master/backend/es6.md#comparison--shortcuts), shortcuts should only be used for `boolean` values. Should be `if (args != null)` here and for all other `null`/`undefined` checks

lib/cli-manager.js
```diff
@@ -0,0 +1,352 @@
...
+const Request = require('./request');
+const utils = require('./utils');
+const cmdInit = require('./commands/init');
+const InitController = require('./controllers/init-controller');
+const cmdFlex = require('./commands/flex');
+const FlexController = require('./controllers/flex-controller');
+const FlexService = require('./services/flex-service');
+
+
+// TODO: refactor the whole setup concept
```
samlevin: I think it's okay to leave TODO statements in the codebase, but if possible, add some context - i.e. add why the setup concept needs to be refactored

lib/authentication.js
```diff
@@ -0,0 +1,97 @@
...
+        });
+
+        done(null, result);
+      }
+    );
+  }
+
+  logout(done) {
+    if (!this.hasCurrentUser()) {
+      return done(null);
```
samlevin: `return done()`

lib/authentication.js
```diff
@@ -0,0 +1,97 @@
...
+      {
+        data,
+        host: host,
+        endpoint: Endpoints.session(),
+        method: HTTPMethod.POST
+      },
+      (err, result) => {
+        debugger;
+        if (err) {
+          // console.log(err);
```
samlevin: Commented out code should be removed (before release)

kinvey.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+const path = require('path');
+// TODO: Figure out if we really need this - seems it's not bad to stay, just remove max upload size
+process.env.NODE_CONFIG_DIR = path.join(__dirname, 'config');
+
+const updateNotifier = require('update-notifier');
+
+const config = require('config');
+const CLIManager = require('./lib/cli-manager');
```
samlevin: Filenames should reflect the name of the default export (https://github.com/Kinvey/style-guides/blob/master/backend/es6.md#naming--filename-matches-export)

Applies to any other instances of the same

kinvey.js
```diff
@@ -0,0 +1,35 @@
...
+ * terms of service and are of legal age to agree to such terms with Kinvey.
+ *
+ * This software contains valuable confidential and proprietary information of
+ * KINVEY, INC and is subject to applicable licensing agreements.
+ * Unauthorized reproduction, transmission or distribution of this file and its
+ * contents is a violation of applicable laws.
+ */
+
+const path = require('path');
+// TODO: Figure out if we really need this - seems it's not bad to stay, just remove max upload size
```
samlevin: Is this going to be addressed as part of this branch? No is fine, but seems like something we should figure out before release

lib/utils.js
samlevin: Commented out lines should be removed

lib/setup.js
samlevin: Comment should be clarified or removed before release

lib/setup.js
samlevin: Is there a ticket for this?

lib/request.js
samlevin: remove debugger everywhere before release

lib/cli-manager.js
DennyGD: I just wanted to give names to the functions executed by async, so that it's clear what's going on.

lib/controllers/flex-controller.js
DennyGD: I've considered it but then I'd have repeating logic for the error handling - try, catch and processing the error. Seems to me it's better to have a `try catch` in the function doing the sync operation and afterwards just check for error. As long as the error handling of sync operations is consistent throughout the code, I think we are fine.

lib/cli-manager.js
DennyGD: Thanks for the link. I really have missed this paragraph   :thumbsup:

lib/setup.js
DennyGD: It's a leftover from a previous state.

lib/controllers/flex-controller.js
samlevin: That might be harder to enforce if different people start to contribute, though, since `try/catch` is well understood but `const err = this.ensureFlexSetup()` is something that we created

lib/cli-manager.js
samlevin: I'd use comments for that purpose then, to avoid saving references to `this` unless it's absolutely needed (for example, in nock): https://github.com/Kinvey/style-guides/blob/master/backend/es6.md#naming--self-this

lib/cli-manager.js
```diff
@@ -0,0 +1,492 @@
...
+
+  /**
+   * Logs out current user if it is a one-time session.
+   * @param done
+   * @returns {*}
+   * @private
+   */
+  _clearSession(done) {
+    if (!this._isOneTimeSession) {
+      return setImmediate(() => { done() });
```
samlevin: Missing semicolon after `done()`

samlevin:
Per previous comment - filenames should reflect the name of the default export (https://github.com/Kinvey/style-guides/blob/master/backend/es6.md#naming--filename-matches-export)

e.g. `cli-manager.js` -> `CLIManager.js` because it exports a class named `CLIManager`

Applies to any other instances of the same

lib/controllers/flex-controller.js
DennyGD: Actually, it is something that I found while thinking what would be a proper solution.

https://stackoverflow.com/questions/7310521/node-js-best-practice-exception-handling/7313005#7313005

https://gist.github.com/balupton/5560110

lib/authentication.js
DennyGD: According to the style guide explicit comparison is required only for strings and numbers.
https://github.com/Kinvey/style-guides/blob/master/backend/es6.md#comparison--shortcuts

Also, the paragraph for checking errors in callbacks uses `if (err)`
https://github.com/Kinvey/style-guides/blob/master/backend/es6.md#requires--check-errors

samlevin:
LGTM, unit tests coming in another branch in this case (per Denny)

PR #52, opened 2017-12-05
Back 2779 integration tests 0

DennyGD:
Integration tests are changed, so that they reflect changes in logic. Now, they test the CLI as black box by spawning a process. Instead of using `nock` (like we did before), a mock server runs on localhost playing the role of MAPI. The output, produced by each command, is saved in a snapshot.
Currently, the integration tests don't include `kinvey init` and `kinvey flex init` (the commands that prompt).
There's some unused or broken code (in `mockServer` and `TestsHelper`) which will be cleared when the unit tests are changed.


tsvetomir-nedyalkov:
LGTM

PR #53, opened 2017-12-12
Overhaul architecture bugfixes

DennyGD:
Fixes and improvements: 

- BACK-3089 Kinvey init: Missing message for invalid credentials
- BACK-3090 Kinvey init: The Profile field should be a required one
- BACK-3091The host url should work without a slash at the end
- BACK-3092 does not work if for host is passed an instance name
- BACK-3118 Kinvey profile list: The command does not show the  existing profiles if KINVEY_CLI_EMAIL and KINVEY_CLI_PASSWORD are set
- BACK-3119 Kinvey profile list: Prettify the output of the command
- BACK-3120 A logout request to MAPI should be made when an overriding/deleting of a profile occurs
- Store job ID in project setup file
- Services by orgId uses correct schema version


lib/CLIManager.js
```diff
@@ -276,6 +270,29 @@ class CLIManager {
...
+   * Makes sure that if profile exists, token will be invalidated. If an error occurs, it logs but doesn't pass it to callback.
+   * @param {String} profileName
+   * @param done
+   * @returns {*}
+   * @private
+   */
+  _logoutUserFromProfile(profileName, done) {
+    const profile = this._setup.findProfileByName(profileName);
+    if (isNullOrUndefined(profile)) {
+      return setImmediate(() => { done(); });
```
dokov: Maybe change this to setImmediate(done)
Also, shouldn't we also include the error?

lib/CLIManager.js
```diff
@@ -409,6 +437,9 @@ class CLIManager {
     let token;
 
     async.series([
+      (next) => {
+        this._logoutUserFromProfile(profileName, next);
```
dokov: I would add a check to make sure the profile existed. Otherwise there would be no need to logout.

DennyGD: the method we're calling here makes the check

lib/CLIManager.js
DennyGD: there would be no error in this case - no job to be done

PR #59, opened 2018-01-08
Dedimitr/2018 sprint 1 bugs and improvements 0

DennyGD:
CLI-23
CLI-25
CLI-9
CLI-19
CLI-10
CLI-7
CLI-18
CLI-4
CLI-3


lib/flex/FlexController.js
```diff
@@ -38,7 +38,7 @@ class FlexController extends BaseController {
     this.flexService = options.flexService;
     this.projectSetup = new ProjectSetup(this.cliManager.config.paths.project);
     this._metadata = {
-      serviceName: 'Service name is not available.',
+      serviceName: '',
```
dokov: It seems more appropriate to set the serviceName to null if you do not have it, instead of an empty string.

lib/flex/FlexService.js
```diff
@@ -168,31 +168,96 @@ class FlexService {
...
+   * Gets all apps and all orgs. If there are no apps and no orgs at the same time, passes error to callback. If entities exist, passes an object with 'app' and 'org' properties.
+   * @param done
+   * @private
+   */
+  _getAllDomainEntities(done) {
+    const result = {};
+
+    async.series([
+      (next) => {
+        this._getDomainEntities(false, (err, data) => {
```
dokov: This does not feel right. The _getDomainEntities function accepts a boolean parameter and uses it to decide whether to return apps or orgs. I think it is better (for readability purposes) to have getApps and getOrgs (they can still call a common function inside). Alternatively, the argument can be changed to a constant instead of bool.

dokov:
And in the future, please do not include so many changes in a single pull request as it is getting really difficult to review. Thanks.

dokov:
LGTM.

PR #64, opened 2018-02-08
Bugfixes and improvements

DennyGD:
Improvements:
* Show connection errors details (CLI-23)
* Show validation errors details (CLI-25)
* Prompt for 'Profile name' instead of 'Profile' (CLI-19)
* Validate 'from' is before 'to' when retrieving logs (CLI-7)
* Prompt for app/org if token is valid (CLI-18)
* Accept '--serviceId' option for 'flex deploy' (CLI-4)
* Enable JSON output format (CLI-28)
* Enable '--no-color' flag (CLI-29)

Bugfixes:
* Don't show flex service name when '--serviceId' (CLI-9)
* Don't invalidate profile token if it can't be overwritten (CLI-10)
* Don't output anything if '--silent' (CLI-36)


dokov:
All commits were reviewed with other pull requests.
LGTM.

PR #66, opened 2018-02-08
Incorporating the latest changes.

boevski:


README.md
```diff
@@ -40,7 +40,7 @@ Kinvey CLI is distributed as an NPM package. After you install NPM, run the foll
 
 * `flex init`
 
-   Configures Kinvey CLI to work with a specific Flex Service through prompts. This command is designed to be executed in a Node.js project directory where it creates a `.kinvey` configuration file. Each successive execution in the same directory overwrites the configuration file. Command-line options that specify the same data as the prompted-for data are ignored.
+   Configures Kinvey CLI to work with a specific Flex Service through prompts. This command is designed to be executed in a Node.js project directory where it creates a `.kinvey` configuration file. Information inside the file is saved per profile. Each successive execution in the same directory overwrites the respective profile section in the configuration file. Command-line options that specify the same data as the prompted-for data are ignored.
```
DennyGD: Add that --email, --password, --instanceId options will be ignored or do you find it redundant?

README.md
```diff
@@ -126,17 +126,17 @@ You can add a global option to every Kinvey CLI command to get the described beh
 
    When used after a `kinvey-cli` command, shows its usage instructions.
 
-* `--host <host>`
+* `--instanceId <instance ID>`
 
-   Hostname of a Kinvey instance. It has a default value of `https://manage.kinvey.com/` which most customers should use. If you are a customer on a dedicated Kinvey instance, enter your dedicated host name as either an absolute URI (`https://kvy-us2-manage.kinvey.com/`) or an instance name only (`kvy-us2`).
+   ID (e.g., `kvy-us2`) or full hostname (e.g., `https://kvy-us2-manage.kinvey.com/`) of a Kinvey instance. It has a default value of `kvy-us1` (or `https://manage.kinvey.com/`) which most customers should use. If you are a customer on a dedicated Kinvey instance, enter your dedicated instance ID.
```
DennyGD: I don't think that clients need to supply full hostname as the instanceId is enough. I'd suggest that we provide info only about the instance ID and skip the hostname part.

README.md
boevski: We've got this sentence at the end to cover just these: "Command-line options that specify the same data as the prompted-for data are ignored." Isn't it good enough?

README.md
boevski: The reason I added this is to let upgrading users get used to the new way; to make it easier for them to understand that instance ID replaces the hostname; to allow a transition period. But I can remove it if required.

README.md
DennyGD: Not in this case because `flex init` doesn't prompt for credentials, anyways. It asks you to choose an app or an org and then, a service from the chosen app or org. But it expects that you are already authenticated. It used to accept email and password as option/env variables but since it starts keeping flex-related info per profile, we need a profile to be used with the command. Either active profile must be set or the command must be invoked with `--profile [profile-name-here]`.

README.md
DennyGD: It makes sense when you put it like that :satisfied: Your call.

README.md
boevski: I see now. An update will follow.

README.md
```diff
@@ -126,17 +126,17 @@ You can add a global option to every Kinvey CLI command to get the described beh
...
 
 * `--no-color`
 
     Disable colors.
     
 * `--output <json>`
```
dokov: This should be "--output <format>" and in the description the allowed values should be listed and explained. The way it is now, it seems I have to write "--output <json>" or "--output <text>"

dokov:
LGTM

PR #72, opened 2018-02-21
CLI-48: Fix flex config; fix logging of errors

DennyGD:


__snapshots__/flex-job.test.js
```diff
@@ -1,6 +1,7 @@
 exports['flex job by specifying a profile and existent jobId should succeed and output default format 1'] = `
 [debug] Checking for package updates
 [debug] Using profile 'profileToGetJobStatus'
+[debug] Error: ENOENT: no such file or directory, open 'projectSetupPath'
```
dokov: This should be more user-friendly. Something like:
Project configuration file not found: ''filepath'

dokov:
Looks good.

PR #75, opened 2018-03-08
Cli 47 add new namespaces

DennyGD:


__snapshots__/coll-create.test.js
```diff
@@ -0,0 +1,215 @@
...
+  --app                     App ID/name                                 [string]
+
+Application is required. Please set active app or use the --app option.
+
+`
+
+exports['coll create with profile when active app is not set without env and without app should fail 1'] = `
+[debug] Checking for package updates
+[debug] Using profile 'activeProfile'
+kinvey coll create <name>
```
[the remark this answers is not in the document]
DennyGD: When help is shown due to a failure, it goes to stderr. When it is shown as a result of running the command with `--help/-h` it goes to stdout. So, we are good 😃

__snapshots__/coll-list.test.js
```diff
@@ -0,0 +1,289 @@
...
+
+
+`
+
+exports['coll list with profile when active app is set active env is not set using existent env name and non-existent app name should fail 1'] = `
+[debug] Checking for package updates
+[debug] Using profile 'activeProfile'
+[debug] Request:  GET http://localhost:3234/v2/apps
+[debug] Response: GET http://localhost:3234/v2/apps 200
+[error] NotFound: Entity not found.
```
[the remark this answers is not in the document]
DennyGD: I thought I have fixed those... apparently not 😄

lib/CommandResult.js
```diff
@@ -65,8 +65,17 @@ class CommandResult {
   _getHumanReadableResult() {
     if (this._humanReadableFormatIsTable()) {
       const tableData = this._getTableData();
-      const formattedTableData = consoleTable.getTable(tableData);
-      return Array.isArray(tableData) ? `Count: ${tableData.length}${EOL}${EOL}${formattedTableData}` : formattedTableData;
+      const isList = Array.isArray(tableData);
+      const sortData = isList && !isEmpty(tableData) && Object.prototype.hasOwnProperty.call(tableData[0], 'name');
```
dokov: I think this code should not be here. If you want something ordered, you should do it in the command handler itself, not here.

DennyGD: We need this for all lists, don't we? This is why I've put it here, it seems common logic to me.

dokov: It is common, but this doesn't mean it belongs here. The normal place to do sorting is where you know what your data is, hence know what to sort it by. What if I sorted my data in the handler and it happens to have a "name" property in the items? How do I sort it by something different?
I think this belongs in a helper method that gets called whenever you need it, but not in the common place where all data goes through.

DennyGD: The moment you need to sort by a different property or not sort at all, you can add a method to the CommandResult that sets that property IMO.
Anyways, I'll add a helper method in Utils.

PR #76, opened 2018-03-11
Qa 163 set travis

tsvetomir-nedyalkov:


__snapshots__/flex-status.test.js
```diff
@@ -163,4 +71,96 @@ exports['flex status by specifying credentials as options when invalid and exist
...
+[debug] Using profile 'profileToGetServiceStatus'
+[debug] Project configuration file not found: 'projectSetupPath'.
+[debug] Request:  GET http://localhost:3234/v2/data-links/12378kdl2/status
+[debug] Response: GET http://localhost:3234/v2/data-links/12378kdl2/status 200
+key            value                                
+-------------  -------------------------------------
+status         ONLINE                               
+version        1.4.2                                
+id             12378kdl2                            
+requestedAt
```
DennyGD: I think it would be better if the date is replaced by something different than an empty string, so that if you take a look at the snapshot you won't think something has gone wrong. It could be 'someDate' or whatever you think suitable.

package.json
```diff
@@ -56,7 +59,7 @@
     "eslint-config-kinvey-platform": "kinvey/eslint-kinvey-platform#v0.1.1",
     "eslint-plugin-import": "2.8.0",
     "express": "4.16.2",
-    "mocha": "2.3.3",
+    "mocha": "^5.0.4",
```
DennyGD: I'd suggest using a specific mocha version.

test/TestsHelper.js
```diff
@@ -393,6 +393,7 @@ TestsHelper.testTooManyArgs = function testTooManyArgs(baseCmd, additionalArgsCo
 TestsHelper.execCmd = function execCmd(cliCmd, options, done) {
   options = options || {
     env: {
+      PATH: process.env.PATH,
```
DennyGD: What's the purpose of this?

test/TestsHelper.js
```diff
@@ -478,6 +479,13 @@ TestsHelper.execCmdWithAssertion = function (cliCmd, cmdOptions, apiOptions, sna
         // ensure line separators are always the same
         finalOutput = finalOutput.replace(/\r\n/g, '\n');
 
+        // regex expression or text to remove from the output
+        if (textToRemove) {
+          const matchedText = finalOutput.match(textToRemove)[0];
+          expect(matchedText).to.exist;
+          finalOutput = finalOutput.replace(matchedText, '');
+        }
```
DennyGD: I can see why you did this but at the same time I feel that the function gets even more complicated. It already accepts too many arguments and is confusing enough. Currently, I can't think of a case where we wouldn't want to replace the date, any date, not just for `flex status`.

test/TestsHelper.js
tsvetomir-nedyalkov: The node variable was not found in Travis - https://travis-ci.org/Kinvey/kinvey-cli/builds/352367292.

DennyGD: Alright. 
Just one more thing - you have to add the `PATH` to the `options`. Currently, if the `options` argument is null or undefined, the `PATH` won't be added and you'd have to add it in multiple places.

test/integration/commands/profile/profile-create.test.js
```diff
@@ -39,6 +39,7 @@ describe('profile create', () => {
   const defaultExpectedSetup = assertions.buildExpectedGlobalSetup({}, expectedProfiles);
 
   const defaultEnv = {
+    PATH: process.env.PATH,
```
DennyGD: This won't be needed if adding the `PATH` happens in one central place.

test/integration/commands/profile/profile-delete.test.js
```diff
@@ -26,6 +26,7 @@ describe('profile delete', () => {
   const defaultProfileName = 'testProfileDelete';
 
   const defaultEnv = {
+    PATH: process.env.PATH,
```
DennyGD: Same here

test/TestsHelper.js
```diff
@@ -554,6 +555,13 @@ TestsHelper.execCmdWithAssertion = function (cliCmd, cmdOptions, apiOptions, sna
         // ensure line separators are always the same
         finalOutput = finalOutput.replace(/\r\n/g, '\n');
 
+        // replace a given text/regex if an object with 'oldValue' and 'newValue' fields is submitted.
+        if (replacementObject && typeof replacementObject === 'object') {
+          const matchedText = finalOutput.match(replacementObject.oldValue)[0];
+          expect(matchedText).to.exist;
```
DennyGD: There's no need to assert that the text is found. If it isn't found, the snapshot will be different.

test/TestsHelper.js
```diff
@@ -460,18 +460,23 @@ TestsHelper.buildCmd = function buildCmd(baseCmd, positionalArgs, options, flags
...
 TestsHelper.execCmd = function execCmd(cliCmd, options, done) {
-  options = options || {
-    env: {
+  options = options || {};
+  if (options.env) {
+    options.env.PATH = options.env.PATH || process.env.PATH;
+    options.env.NODE_CONFIG = options.env.NODE_CONFIG || JSON.stringify(testsConfig);
+  } else {
+    options.env = {
+      PATH: process.env.PATH,
```
DennyGD: I think it should be more like:
- whether or not options are null or not => set `PATH` (because of travis and add comment)
- if options are null => set `NODE_CONFIG`
- if options are not null => do not set `NODE_CONFIG`

And now options may not be null and `NODE_CONFIG` may intentionally be not set but this will set it.

DennyGD:
LGTM

PR #78, opened 2018-03-22
Cli 52 use active profile

DennyGD:
Use the active profile (if set) if no profile name is supplied for `kinvey profile delete`

### DIFF
--- a/__snapshots__/profile-delete.test.js
+++ b/__snapshots__/profile-delete.test.js
@@ -44,27 +44,18 @@ exports['profile delete by non-existent name when none should return error 1'] =
 
 `
 
-exports['profile delete without a name should fail 1'] = `
-kinvey profile delete <name>
+exports['profile delete without a name when no active should fail 1'] = `
+[debug] Checking for package updates
+[error] ItemNotSpecified: No profile identifier is specified and/or active profile is not set.
 
-Delete profile by name
+`
 
-Positionals:
-  name  Profile name                                                  [required]
-
-Options:
-  --version                 Show version number                        [boolean]
-  --email                   E-mail address of your Kinvey account       [string]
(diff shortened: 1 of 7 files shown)


__snapshots__/profile-delete.test.js
```diff
@@ -44,27 +44,18 @@ exports['profile delete by non-existent name when none should return error 1'] =
...
-exports['profile delete without a name should fail 1'] = `
-kinvey profile delete <name>
-
-Delete profile by name
-
-Positionals:
-  name  Profile name                                                  [required]
+exports['profile delete without a name when no active should fail 1'] = `
+[debug] Checking for package updates
+[error] ItemNotSpecified: No profile identifier is specified and/or active profile is not set.
```
dokov: Isn't this just "and" not "and/or"?

lib/profile/profilesRouter.js
```diff
@@ -49,8 +49,8 @@ const cmdDefinitions = [
     handlerName: 'use'
   },
   {
-    command: 'delete <name>',
-    desc: 'Delete profile by name',
+    command: 'delete [name]',
+    desc: 'Delete a specified profile or the active one',
```
dokov: How about: "... the active one if no profile is specified"

__snapshots__/profile-delete.test.js
DennyGD: It is, indeed 😄

lib/profile/profilesRouter.js
DennyGD: I like it better like this. It's not focused so much on the active profile and it just gives you all the options. But it's up to you?

dokov: OK, leave it like that then.

PR #80, opened 2018-03-27
Cli 53 password command

DennyGD:


lib/profile/ProfilesController.js
```diff
@@ -157,6 +159,53 @@ class ProfilesController extends BaseController {
...
+    }
+
+    if (!this.cliManager.profileExists(profileName)) {
+      const err = new KinveyError(Errors.ProfileNotFound);
+      return done(err);
+    }
+
+    async.waterfall([
+      (next) => {
+        const q = this.cliManager.prompter.buildQuestion(
```
dokov: I am not sure what it displays exactly, but I think we should have the email in the prompt somewhere.
Something like:
Email: email@example.com
Password:

DennyGD: We already have it in the profile :confused:

lib/profile/profilesRouter.js
```diff
@@ -56,6 +56,15 @@ const cmdDefinitions = [
         .positional('name', { describe: 'Profile name' });
     },
     handlerName: 'deleteProfile'
+  },
+  {
+    command: 'login [name]',
+    desc: 'Acquire new authentication token',
```
dokov: I think "tokens" is not something that the user knows or cares about. It should be something like:
"Re-authenticate a profile. If no profile is specified, the active profile is used."
Similarly, we need to change the description of the "show" command:
"View detailed info for a profile. If no profile is specified, the active profile is used.

lib/profile/ProfilesController.js
dokov: I mean, not ask for it, but show it, so that the customer knows which password he needs to write.

PR #81, opened 2018-03-28
Cli 55 env config

DennyGD:
Basic functionality for creating and updating environments + their collections

### DIFF
--- a/lib/CLIManager.js
+++ b/lib/CLIManager.js
@@ -55,6 +55,8 @@ const ApplicationsService = require('./application/ApplicationsService');
 const EnvironmentsService = require('./environment/EnvironmentsService');
 const CollectionsService = require('./collection/CollectionsService');
 
+const ConfigFileProcessor = require('./ConfigFileProcessor');
+
 /**
  * Serves as mediator between modules. Responsible for routing and controllers initializing.
  */
@@ -90,6 +92,14 @@ class CLIManager {
     this._controllers[Namespace.ENV] = new EnvironmentsController({ cliManager: this, environmentsService, applicationsService });
     const collectionsService = new CollectionsService(this);
     this._controllers[Namespace.COLL] = new CollectionsController({ cliManager: this, collectionsService, environmentsService });
+
+    this._configFileProcessor = new ConfigFileProcessor({
+      cliManager: this,
+      organizationsService,
+      applicationsService,
+      environmentsService,
+      collectionsService
+    });
   }
 
   getController(name) {
@@ -775,6 +785,10 @@ class CLIManager {
 
       done(err);
     });
+  }
+
+  processConfigFile(operationType, options, done) {
+    this._configFileProcessor.process(operationType, options, done);
   }
 }
 
--- a/lib/ConfigFileProcessor.js
+++ b/lib/ConfigFileProcessor.js
@@ -15,7 +15,7 @@
 
 const async = require('async');
 
-const { AppOptionsName, DomainTypes, EntityType, EnvOptionsName, OperationType } = require('./Constants');
+const { AppOptionsName, DomainTypes, EntityType, EnvOptionsName, LogLevel, OperationType } = require('./Constants');
 const { Endpoints, getConfigTypeError, getCustomNotFoundError, getObjectByOmitting, isEmpty, isNullOrUndefined, readJSON } = require('./Utils');
 
 class ConfigFileProcessor {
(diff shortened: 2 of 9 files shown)


lib/environment/EnvironmentsController.js
```diff
@@ -55,10 +55,19 @@ class EnvironmentsController extends BaseController {
...
-          name: options.name
-        };
-        this.environmentsService.create(newEnv, appId, next);
+        if (options.file) {
+          const info = {
+            [AppOptionsName.APP]: appId,
+            file: options.file,
+            name: options.name
+          };
+          this.cliManager.processConfigFile(OperationType.CREATE, info, next);
```
dokov: The "info" variable is called "options" in the processConfigFile method. I see that you already have "options" here, but you should probably call this "processOptions" or something like that. "info" does not have any meaning.

lib/environment/EnvironmentsController.js
```diff
@@ -73,6 +82,29 @@ class EnvironmentsController extends BaseController {
...
+    this.environmentsService.getActiveOrSpecified(options, (err, env) => {
+      if (err) {
+        return done(err);
+      }
+
+      const info = {
+        [EnvOptionsName.ENV]: env,
+        file: options.file
+      };
+      this.cliManager.processConfigFile(OperationType.UPDATE, info, (err, data) => {
```
dokov: Same as above on the "info" and "options".

lib/ConfigFileProcessor.js
```diff
@@ -0,0 +1,327 @@
...
+  constructor(options) {
+    this.cliManager = options.cliManager;
+    this.organizationsService = options.organizationsService;
+    this.applicationsService = options.applicationsService;
+    this.environmentsService = options.environmentsService;
+    this.collectionsService = options.collectionsService;
+  }
+
+  _create(source, options, done) {
+    if (source.configType === 'environment') {
```
dokov: The idea of configType is not to give you a way to fork your code. The idea is to be able to make certain that the user supplied the proper config file for the command. For example, you should not be able to give an environment config file to "kinvey app create".
I can see that the code will throw an error anyway, but we need to give the precise error, something like:
"You have specified the wrong type of config file. Expected: application"

lib/CLIManager.js
```diff
@@ -777,6 +787,10 @@ class CLIManager {
       done(err);
     });
   }
+
+  processConfigFile(operationType, options, done) {
```
dokov: I think that here you should either put the operationType inside the options object, or get everything out of the options object.

Also, if you decide to have the options object, I would change the ID of the object to always be "objectId", and not differ based on what type of configuration you are currently using it for (e.g. appId, envId, etc.)

Alternatively, you can have two separate methods. This is what I would do, as you do not seem to share any code between update and create, aside from the JSON parsing. Here are the methods:
processCreate(objectType, name, configFile);
processUpdate(objectType, objectId, configFile);

lib/ConfigFileProcessor.js
```diff
@@ -0,0 +1,327 @@
...
+ * Unauthorized reproduction, transmission or distribution of this file and its
+ * contents is a violation of applicable laws.
+ */
+
+const async = require('async');
+
+const { AppOptionsName, DomainTypes, EntityType, EnvOptionsName, LogLevel, OperationType } = require('./Constants');
+const { Endpoints, getConfigTypeError, getCustomNotFoundError, getObjectByOmitting, isEmpty, isNullOrUndefined, readJSON } = require('./Utils');
+
+class ConfigFileProcessor {
```
dokov: The way it is, this file will grow a lot with the implementation of more config file features. I would separate it for the different types of configuration - environment, service, etc.

lib/CLIManager.js
DennyGD: I'd prefer keeping the `options` object and have it include the operationType.
As for the `objectId` - I'd rather not because I think that for different operations/configTypes I'd need different things and not neccesarily just one id.

lib/ConfigFileProcessor.js
DennyGD: Yeah, I agree. I was just thinking of doing the separation when the need arises as then I'd have more clarity how exactly to separate things.

lib/CLIManager.js
dokov: What different things do you expect to need for the different operations?

lib/ConfigFileProcessor.js
dokov: That's fine, but we will definitely need to do this before MVP.

lib/CLIManager.js
DennyGD: In order to update the env the whole env object is needed. For creating dataLinks(services) we will probably need an appId or orgId and we will need to know whether it is an app or org id.
I feel that the file processor have to receive all the data it needs and to focus on processing the config files.

dokov: The way I see it, we will not need any additional info. Aside from the name (which will later be the slug, once we have slugs), everything will be in the config file.
But OK, let's leave it like that for now. We'll see how it looks when we add another type of config file (service) and will change it if necessary.

PR #86, opened 2018-04-16
CLI-70: Overwrite only profile token on 'profile login'

DennyGD:


__snapshots__/profile-login.test.js
```diff
@@ -0,0 +1,17 @@
...
+
+exports['profile login without active profile when several profiles and none specified should fail 1'] = `
+[debug] Checking for package updates
+[error] ItemNotSpecified: No profile identifier is specified and active profile is not set.
+
+`
+
+exports['profile login without active profile when several profiles and non-existent specified should fail 1'] = `
+[debug] Checking for package updates
+[error] ProfileNotFound: Profile not found. Please verify profile name exists.
```
dokov: I think we should also show the name of the profile that does not exist.

DennyGD: Well, it's the name that the user has just specified...

PR #87, opened 2018-04-19
CLI-62: Update README

DennyGD:


README.md
```diff
@@ -12,7 +12,13 @@
 
 # Kinvey CLI
 
-Kinvey CLI is a utility for deploying and managing FlexServices running on the Kinvey FlexService Runtime.
+Kinvey CLI is a utility for managing various aspects of your Kinvey account from the command line. Its features include:
+
+* Deploying and managing FlexServices running on the Kinvey FlexService Runtime
+* Viewing organizations and setting a default organization
```
DennyGD: Setting default/active organization is in the scope of the CLI only. Seeing it mentioned here makes me think that it's a change that happens in the backend. Not sure if it is just me...

README.md
```diff
@@ -54,9 +60,109 @@ Kinvey CLI is distributed as an NPM package. After you install NPM, run the foll
...
+
+    Lists all existing applications within your Kinvey account.
+    
+* `app show [app]`
+
+    Shows detailed information about the specified application or about the active one if you don't specify an application. You can specify an application by ID or name.
+
+* `app use <app>`
+
+    Sets the specified application as active.
```
DennyGD: Not saying that it should be included here but I just want to mention that the app can again be specified by ID or name. Same applies for `org use <org>`.

README.md
```diff
@@ -54,9 +60,109 @@ Kinvey CLI is distributed as an NPM package. After you install NPM, run the foll
...
+       
+      Specifies a Kinvey app by ID or name. Requires `--env`.
+    
+* `env show [env]`
+
+    Shows detailed information about the specified environment or about the active one if you don't specify an environment. You can specify an environment by ID or name. By default, the command searches inside the active application but you can specify a different application using `--app`.
+    
+    * `--app <application>`
+       
+      Specifies a Kinvey app by ID or name. Requires `--env`.
```
DennyGD: Actually, it requires `--env` in the collections context. Here, the app only is enough. The same applies for the entire namespace.

README.md
boevski: You are right. Organization selection only makes sense under Managing Flex so I'm deleting this bullet.

README.md
boevski: Included.

README.md
boevski: Thanks, that's a copying-and-pasting error.

PR #97, opened 2018-06-13
Cli 81 config file validation

DennyGD:
Validate env and service config files


__snapshots__/schema-validator.test.js
```diff
@@ -0,0 +1,19 @@
+exports['schema validator env with invalid env should fail 1'] = `
+
+	version: "version" must be one of [1.0.0]
```
dokov: I think this should be "schemaVersion", not simply "version". "Version" might confuse customers that it is meant to hold the version of the backend configuration file.

PR #105, opened 2018-06-22
CLI-82: Small fixes

dokov:


lib/exporter/EnvFileExporter_1_0_0.js
```diff
@@ -70,7 +69,7 @@ class EnvFileExporter_1_0_0 {
         });
       },
       (next) => {
-        this._exportSettings(env, (err, settings) => {
+        EnvFileExporterV1._exportSettings(env, (err, settings) => {
```
DennyGD: Err handling?

lib/exporter/EnvFileExporter_1_0_0.js
```diff
@@ -362,29 +353,27 @@ class EnvFileExporter_1_0_0 {
     });
   }
 
-  _mustExportCollection(collectionName) {
+  static _mustExportCollection(collectionName) {
```
DennyGD: What is the idea behind this method?

lib/exporter/EnvFileExporter_1_0_0.js
```diff
@@ -397,21 +386,20 @@ class EnvFileExporter_1_0_0 {
...
-
             const configPermissionsById = {};
-            self._transformPermissionsObject(configPermissionsById, apiPermissions, 'create');
-            self._transformPermissionsObject(configPermissionsById, apiPermissions, 'read');
-            self._transformPermissionsObject(configPermissionsById, apiPermissions, 'update');
-            self._transformPermissionsObject(configPermissionsById, apiPermissions, 'delete');
+            EnvFileExporterV1._transformPermissionsObject(configPermissionsById, apiPermissions, 'create');
+            EnvFileExporterV1._transformPermissionsObject(configPermissionsById, apiPermissions, 'read');
+            EnvFileExporterV1._transformPermissionsObject(configPermissionsById, apiPermissions, 'update');
+            EnvFileExporterV1._transformPermissionsObject(configPermissionsById, apiPermissions, 'delete');
```
DennyGD: Constants.AllowedPermissionsOperations

lib/exporter/EnvFileExporter_1_0_0.js
```diff
@@ -15,15 +15,16 @@
...
+const findKey = require('lodash.findkey');
 
 
 const Utils = require('../Utils');
 
 const { ActiveItemType, AppOptionsName, CommonOptionsNames, ConfigType, EntityType, EnvOptionsName, Errors, Mapping, OperationType, BackendCollectionPermission, DomainTypes } = require('../Constants');
 const KinveyError = require('../KinveyError');
 
-class EnvFileExporter_1_0_0 {
+class EnvFileExporterV1 {
```
DennyGD: File should be renamed too -> https://github.com/Kinvey/style-guides/blob/master/backend/es6.md#naming--filename-matches-export

lib/exporter/EnvFileExporter_1_0_0.js
```diff
@@ -453,17 +441,17 @@ class EnvFileExporter_1_0_0 {
       hookObject.type = 'internal';
       let filename;
       if (this.storeCodeAsFile) {
-        filename = 'hooks/' + apiCollectionHook.name + '.json';
+        filename = 'hooks/' + apiCollectionHook.name + '.js';
```
DennyGD: String interpolation is much nicer IMO
https://github.com/Kinvey/style-guides/blob/master/backend/es6.md#es6-template-literals

lib/exporter/EnvFileExporter_1_0_0.js
```diff
@@ -147,26 +146,25 @@ class EnvFileExporter_1_0_0 {
...
+        this.collectionsService.getAll(envId, next);
       },
       (collections, next) => {
         async.forEachSeries(
           collections,
           (collection, iter) => {
-            if (self._mustExportCollection(collection.name)) {
-              self._transformCollection(collection, apiRoles, (err, configCollection) => {
+            if (EnvFileExporterV1._mustExportCollection(collection.name)) {
+              this._transformCollection(collection, apiRoles, (err, configCollection) => {
```
DennyGD: Err handling?

DennyGD:
Apart from the other comments I have a few more which are for unchanged lines but still relevant:
* EnvironmentsController, EnvFileExporterV1 - `export` is a reserved word
* When calling `Utils.mkdirp` the exception should be handled, otherwise it gets a bit ugly

lib/exporter/EnvFileExporter_1_0_0.js
```diff
@@ -53,15 +52,15 @@ class EnvFileExporter_1_0_0 {
       (next) => {
         this.applicationsService.getByIdOrName(env.app, (err, app) => {
           if (err) return next(err);
-          self.servicesService.getAllByDomainType(DomainTypes.APP, { id: env.app }, (errAppServices, appServices) => {
+          this.servicesService.getAllByDomainType(DomainTypes.APP, { id: env.app }, (errAppServices, appServices) => {
```
DennyGD: I'm probably missing something but why do we need all the services?

dokov: We need them for the export of custom endpoints and collection hooks. There we only have the ID of the backing server that handles the endpoint/hook and we want to export the name of the service. In order to get the name of the service, we need to get all available services and then search the one with the specific backing server ID. So, in order to do that I get all the services once, in the beginning, instead of getting them for every endpoint/hook I want to export.

lib/exporter/EnvFileExporter_1_0_0.js
dokov: Good catch, will fix.

lib/exporter/EnvFileExporter_1_0_0.js
dokov: Good catch. Agree on both.

lib/exporter/EnvFileExporter_1_0_0.js
dokov: Well, I hoped to leave the file like that, but since it is official, I will rename it as well.

Initially I thought of the schema version as a "1.0.0" type of versioning. While renaming the classes, I thought that we might want to just have a single number version for schema version.

DennyGD:
I still feel that we shouldn't be using reserved words for naming functions. Other than that, it LGTM

PR #106, opened 2018-06-25
CLI-88: Support groups in env config file

DennyGD:


README.md
```diff
@@ -689,6 +700,14 @@ and **accessType** is one of: *never, always, grant, entity*
 
 `roles.[roleName].description` Role description. String. Optional.
 
+`groups` Groups. Optional. Object where each first-level property is the ID of a group.
```
dokov: Maybe add an example here, as currently the ID is supposed to be user friendly...
Something like: If your name is "My Group", your ID should be "MyGroup" or "my-group".

PR #107, opened 2018-06-25
CLI-98: Fixed export of push settings. Removed empty objects from exported config.

dokov:


lib/exporter/EnvFileExporterV1.js
```diff
@@ -359,7 +368,14 @@ class EnvFileExporterV1 {
       },
       (pushSettings, callback) => {
         // Remove iOS push settings - we do not support those currently
```
DennyGD: Comment does not apply anymore, does it?

DennyGD:
LGTM

PR #108, opened 2018-06-29
QA-228: Adjust linting to be consistent with rest of the Kinvey components

dimofeev:
Changes:
 - add eslintignore file;
 - update npm scripts;
 - remove the custom lint rules;
 - update the Copyright date.


DennyGD:
Tests have failed.

dimofeev:
Fixed. The problem was in the BIN folder. In one of the latest versions of eslint, a breaking change was introduced that require .js files in the folders selected for scan. As per Denny's request, the BIN folder was removed from the script for linting.

.eslintrc
```diff
@@ -1,13 +1,3 @@
 {
-  "extends": "eslint-config-kinvey-platform",
-  "env": {
```
DennyGD: Is there any problem in specifying the environment? After all, it's nothing more than just pointing eslint to what already is.

.eslintrc
```diff
@@ -1,13 +1,3 @@
...
-    "mocha": true,
-    "node": true
-  },
-  "globals": {
-    "expect": true
```
DennyGD: It seems common to define `expect` as a global variable and skip the need to require it in each and every test file. The latter can easily become annoying IMO. What's your take on this?

DennyGD:
I have a few concerns here. 
First, I'm not entirely sure why we want to stick with the `no-param-reassign` rule as it is common to have default values for function parameters and write something like this: 
```
function doSomething(a, b) {
  a = a || 'someDefaultValue';
  ...
}
```
But even if we do, this will require code changes in order to achieve this immutability.
I'm quite hesitant about merging these changes into master as the tests are failing and until the code base is fixed we'll have them failing. Other errors could easily slip by, too. 

@dokov what do you think?

.eslintignore
```diff
@@ -0,0 +1 @@
+node_modules/*
```
DennyGD: They are ignored by default...

dimofeev: That's true, but we can always add new files/folders to it when needed. That's the reason for including it there - avoiding empyt file.

.eslintrc
DennyGD: https://eslint.org/docs/user-guide/configuring#specifying-environments

.eslintrc
DennyGD: I think it is required there but still, if not declared as global, eslint complains.

.eslintrc
dimofeev: Does it cause any errors?

.eslintrc
dimofeev: I cannot see it there. What happens when we set it there - still eslint errors?

DennyGD: It requires `common.js` and inside this file `expect` is required.

.eslintrc
DennyGD: I don't know about the errors...
My point is that this project is meant to be run in node.js environment and it contains mocha tests. Specifying this `env` object only indicates to eslint the status quo.

dimofeev:
The problem with global variables is fixed within new versio, resulting in 160 errors less. Hence 30 errors left. Please, adjust you code to pass these and merge the PR. Thanks.

DennyGD:
Commits will be merged alongside the lint errors fixes in a separate PR.

PR #117, opened 2018-07-25
Cli 95 flex service export tests

DennyGD:
Add tests communicating with the backend on production for:
* Creating an environment from a config file
* Modifying an env from a config file
* Exporting an env to a config file
* Creating a flex service from a config file
* Modifying a flex service from a config file
* Exporting a flex service to a config file


tsvetomir-nedyalkov:
It is good to have some sort of a guide how to run the tests against different environments, which explains for example that KINVEY_CLI_INSTANCE_ID should be set and what existing Backend entities are needed.

tsvetomir-nedyalkov:
When the tests are run against the default environment without setting KINVEY_CLI_INSTANCE_ID, the instance is shown as undefined in the logs - it is good to show the current environment.

test/integration-no-mock/config-management/env-config-create.js
```diff
@@ -0,0 +1,404 @@
...
+      (next) => {
+        EnvHelper.assertEnvOnly(env, envId, envName, next);
+      },
+      (next) => {
+        EnvHelper.assertCollections(envId, collList, collList.length + 2, null, next);
+      }
+    ], done);
+  });
+
+  it('internal collections only plus system collections should succeed', (done) => {
```
tsvetomir-nedyalkov: If the purpose of the test if to verify that user and _blob collections will be updated, we should pass a different type of permission for them - e.g. 'private'. Otherwise, currently the test will pass even if _blob and user collections are not updated from the cli.

test/ApiService.js
```diff
@@ -0,0 +1,322 @@
...
+
+let authTokenMapi;
+const baasAuthTokens = {};
+
+function getBaseUrl(isBaasRequest) {
+  if (isBaasRequest && process.env.KINVEY_CLI_BAAS) {
+    return process.env.KINVEY_CLI_BAAS;
+  }
+
+  let instance = process.env.KINVEY_CLI_INSTANCE;
```
tsvetomir-nedyalkov: It seems that the correct variable is KINVEY_CLI_INSTANCE_ID. Otherwise the tests cannot be run on a different environment.

test/integration-no-mock/config-management/env-config-create.js
```diff
@@ -0,0 +1,404 @@
...
+        EnvHelper.assertCollections(envId, collList, collList.length, rolesNameIdPairs, next);
+      },
+      (next) => {
+        EnvHelper.assertAllCollHooks(envId, collHooks, next);
+      }
+    ], done);
+  });
+
+  it('common code, internal endpoints, groups and push settings should succeed', (done) => {
+    const commonCode = {
```
tsvetomir-nedyalkov: The test does not assert that the common code is created correctly and will not fail if it isn't.

test/integration-no-mock/config-management/env-config-export.js
```diff
@@ -0,0 +1,130 @@
...
+          if (err) {
+            return next(err);
+          }
+
+          exportedConfig = result;
+          next();
+        });
+      },
+      (next) => {
+        ConfigManagementHelper.env.assertExportedConfig(envConfig, exportedConfig, next);
```
tsvetomir-nedyalkov: 'exportedConfig' has a 'settings' property, which is not validated

test/integration-no-mock/config-management/env-config-export.js
```diff
@@ -0,0 +1,130 @@
...
+
+  before('setup app', (done) => {
+    ConfigManagementHelper.app.createInTestsOrg({ name: appName }, done);
+  });
+
+  after('remove app', (done) => {
+    execCmdWoMocks(`app delete ${appName} --no-prompt`, null, done);
+  });
+
+  it('with collections, hooks, endpoints, push settings, groups, roles should succeed', (done) => {
```
tsvetomir-nedyalkov: The test does not fail if the custom endpoints, push settings, groups or roles are not exported correctly. It seems that the problem is in the used assertion function - _ConfigManagementHelper.env.assertExportedConfig_

test/ConfigManagementHelper.js
```diff
@@ -0,0 +1,715 @@
...
+          expect(actual.host).to.not.be.null;
+        }
+
+        const defaultCode = 'function onRequest(request, response, modules) {\n  response.continue();\n}';
+        const expectedCode = expected.code || defaultCode;
+        expect(expectedCode).to.equal(actual.code);
+
+        if (expected.schedule) { // due to api peculiarities
+          if (!expected.schedule.interval) {
+            expect(expected.schedule.start).to.equal(actual.schedule);
```
tsvetomir-nedyalkov: 'actual' and 'expected' places should be exchanged. The same is valid for the assertions in the _else_ clause.

test/ConfigManagementHelper.js
```diff
@@ -0,0 +1,715 @@
...
+        const expectedHook = collHooks[currentHook];
+        if (expectedHook.type === 'internal') {
+          expect(actualHook.host).to.be.null;
+        } else {
+          expect(actualHook.host).to.not.be.null;
+        }
+
+        const defaultCode = `function ${currentHook}(request, response, modules) {\n  response.continue();\n}`;
+        const expectedCode = expectedHook.code || defaultCode;
+        expect(expectedCode).to.equal(actualHook.code);
```
tsvetomir-nedyalkov: 'actual' and 'expected' places should be exchanged.

test/ConfigManagementHelper.js
```diff
@@ -0,0 +1,715 @@
...
+  return expectedPermissionsPerColl;
+};
+
+env.assertCollections = function assertCollections(envId, collList, expCollCount, rolesNameIdPairs, done) {
+  ApiService.colls.get(envId, null, (err, actualColls) => {
+    if (err) {
+      return done(err);
+    }
+
+    expect(expCollCount).to.equal(actualColls.length);
```
tsvetomir-nedyalkov: 'actual' and 'expected' places should be exchanged. There are several places in this file with this problem - can be found with search by 'expect(exp'.

test/integration-no-mock/config-management/env-config-push.js
```diff
@@ -0,0 +1,172 @@
...
+
+  before('setup app', (done) => {
+    ConfigManagementHelper.app.createInTestsOrg({ name: appName }, done);
+  });
+
+  after('remove app', (done) => {
+    execCmdWoMocks(`app delete ${appName} --no-prompt`, null, done);
+  });
+
+  it('modifying collections, groups, endpoints and push settings; creating hooks; omitting roles should succeed', (done) => {
```
tsvetomir-nedyalkov: It seems that the _groups_ update is not verified by the test

test/integration-no-mock/config-management/env-config-push.js
```diff
@@ -0,0 +1,172 @@
...
+        ConfigManagementHelper.env.assertEnvOnly(modifiedEnv, envId, envName, next);
+      },
+      (next) => {
+        ApiService.roles.get(envId, null, (err, actualRoles) => {
+          if (err) {
+            return next(err);
+          }
+
+          const expectedRolesCount = rolesList.length;
+          expect(expectedRolesCount).to.equal(actualRoles.length);
```
tsvetomir-nedyalkov: Expected and Actual should be exchanged.

test/integration-no-mock/config-management/env-config-create.js
```diff
@@ -0,0 +1,404 @@
...
+      (next) => {
+        ApiService.roles.get(envId, null, (err, data) => {
+          if (err) {
+            return next(err);
+          }
+
+          actualRoles = data;
+
+          const expectedRolesCount = rolesList.length;
+          expect(expectedRolesCount).to.equal(actualRoles.length);
```
tsvetomir-nedyalkov: Expected and Actual should be exchanged.

test/integration-no-mock/config-management/service-config-create.js
```diff
@@ -0,0 +1,147 @@
...
+
+const async = require('async');
+const moment = require('moment');
+const path = require('path');
+
+const ApiService = require('./../../ApiService');
+const ConfigManagementHelper = require('./../../ConfigManagementHelper');
+
+const { randomStrings } = require('./../../TestsHelper');
+const { writeJSON } = require('../../../lib/Utils');
```
tsvetomir-nedyalkov: An unused variable

test/integration-no-mock/config-management/service-config-push.js
```diff
@@ -0,0 +1,218 @@
...
+      (next) => {
+        ConfigManagementHelper.service.modifyFromConfig(serviceId, modifiedServiceConfig, null, next);
+      },
+      (next) => {
+        ConfigManagementHelper.service.assertFlexService(serviceId, modifiedServiceConfig, serviceName, next);
+      }
+    ], done);
+  });
+
+  it('modifying only description on internal flex service with sourcePath should succeed', (done) => {
```
tsvetomir-nedyalkov: This test is a duplicate of the previous one

test/ConfigManagementHelper.js
```diff
@@ -0,0 +1,715 @@
...
+    if (err) {
+      return done(err);
+    }
+
+    done(null, results.pop());
+  });
+}
+
+const env = {};
+env.buildSettings = function buildSettings(options) {
```
tsvetomir-nedyalkov: _options_ are not used

test/ApiService.js
```diff
@@ -0,0 +1,322 @@
...
+  if (instance && instance.includes('localhost')) {
+    return instance;
+  }
+
+  if (!instance) {
+    instance = 'kvy-us1';
+  }
+
+  return isBaasRequest ? `https://${instance}-baas.kinvey.com` : `https://${instance}-manage.kinvey.com`;
```
tsvetomir-nedyalkov: Is _getBaseUrl_ working on the local dev machine - isn't the MAPI Url something like http://localhost:8888?

test/integration-no-mock/config-management/env-config-create.js
DennyGD: It is more to ensure that all internal collections will be created correctly, that error won't occur because of system collections ('cause unlike other collections they should be updated, not created) and that there will be no errors just because we have only collections (not a full config). It's just a base case, really.

test/integration-no-mock/config-management/env-config-push.js
DennyGD: I caught myself several times while writing the tests thinking which goes first and which second :grin: Will fix.

test/integration-no-mock/config-management/env-config-create.js
DennyGD: Nice catch :clap:

DennyGD:
Concerning the guide to run the tests: I think that we should have one, too. I'm not sure though whether it should be in the README in the repo or in a wiki. We already use the README to document the CLI and I'm reluctant to add even more info there and moreover such that does not concern the clients. On the plus side, it's easier for a dev/QA to find info if it's all in the repo. Still, I'm more inclined to write a wiki. There are some peculiarities when the backend runs on localhost that are really redundant for an outsider.

test/integration-no-mock/config-management/env-config-export.js
tsvetomir-nedyalkov: The roles are still not validated

DennyGD: Yes, this is intentional. I've found a bug and logged it (cli-135). Until it is fixed, they will fail and I decided to skip them and add a TODO there.

tsvetomir-nedyalkov:
LGTM

PR #118, opened 2018-08-06
Flex-198: Extended flex status command to show deployment information

galexandrov:
The deployment status will be shown when `flex-operator` feature falg is enabled for application

Flex status command output

```
key                value                                
-----------------  -------------------------------------
status             ONLINE                               
version            1.4.2                                
runtime            FLEX-RUNTIME-NODE8                   
id                 123456                               
requestedAt        replaced_value
deployerEmail      davy.jones@mail.com                  
deployerName       Davy Jones                           
deploymentStatus   COMPLETED                            
deploymentVersion  1.4.2                                
deploymentRuntime  FLEX-RUNTIME-NODE10  
```

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -227,6 +227,12 @@ Constants.ServiceStatus = {
   ERROR: 'ERROR'
 };
 
+Constants.DeploymentStatus = {
+  RUNNING: 'RUNNING',
+  ERROR: 'ERROR',
+  COMPLETED: 'COMPLETED'
+};
+
 Constants.FlexProjectMaxSize = 10 * 1024 * 1024;
 
 /**
@@ -413,9 +419,7 @@ Constants.Mapping[Constants.EntityType.APP] = {
     id: 'id',
     name: 'name',
     organizationId: 'organizationId',
-    environments: (item) => {
-      return item && item.environments ? item.environments.length : 0;
-    }
+    environments: item => (item && item.environments ? item.environments.length : 0)
   }
 };
 Constants.Mapping[Constants.EntityType.APP].DETAILS = Object.assign(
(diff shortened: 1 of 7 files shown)


test/fixtures/datalinks-status-response.json
```diff
@@ -0,0 +1,78 @@
...
+      "deployUserInfo": {
+        "firstName": "Davy",
+        "lastName": "Jones",
+        "email": "davy.jones@mail.com"
+      },
+      "version": "1.4.2",
+      "runtime": "flex-runtime-node8",
+      "deployment": {
+        "status": "COMPLETED",
+        "version": "1.4.2",
```
tsvetomir-nedyalkov: It is good 'version' to be different from 'deployment.version' in order to validate that the CLI reads and displays the correct fields from MAPI

test/fixtures/datalinks-status-response.json
```diff
@@ -0,0 +1,78 @@
...
+      "requestedAt": "2017-11-06T03:42:31.970Z",
+      "deployUserInfo": {
+        "firstName": "Davy",
+        "lastName": "Jones",
+        "email": "davy.jones@mail.com"
+      },
+      "version": "1.4.2",
+      "deployment": {
+        "status": "COMPLETED",
+        "version": "1.4.2",
```
tsvetomir-nedyalkov: Same as above

lib/Constants.js
```diff
@@ -227,6 +227,12 @@ Constants.ServiceStatus = {
   ERROR: 'ERROR'
 };
 
+Constants.DeploymentStatus = {
```
tsvetomir-nedyalkov: It seems that these Constant is not used anywhere...

galexandrov: https://github.com/Kinvey/kinvey-cli/blob/flex-198-operator-integration/lib/flex/FlexService.js#L34

PR #119, opened 2018-08-06
CLI-84: Implemented file references

dokov:


dokov:
Agreed. Resolving logic moved to ConfigFileProcessor.

lib/ConfigFileProcessor.js
```diff
@@ -75,6 +80,79 @@ class ConfigFileProcessor {
...
     });
   }
+
+  /**
+   * Reads a JSON configuration file and returns a JavaScript object. Does all the necessary processing and resolves all referenced files.
+   * @param {string} filePath
+   */
+  static readJSONConfigFile(file, cb) {
+    ConfigFileProcessor._readJSONConfigFile(file, 'root', (err, config) => {
+      logger.debug('Final configuration: ' + JSON.stringify(config, null, 4));
```
DennyGD: What if there is an error here?

lib/ConfigFileProcessor.js
```diff
@@ -75,6 +80,79 @@ class ConfigFileProcessor {
...
+      }
+    ], cb);
+  }
+
+  static _processConfigFileValue(obj, baseFolder, fieldPath, done) {
+    if (Array.isArray(obj) || lodashIsObject(obj)) {
+      async.forEachOfSeries(
+        obj,
+        (value, key, callback) => {
+          // console.log(key + '    ' + JSON.stringify(value));
```
DennyGD: Forgotten, I guess

lib/ConfigFileProcessor.js
```diff
@@ -75,6 +80,79 @@ class ConfigFileProcessor {
...
+          logger.debug('Referenced files resolved for %s', chalk.cyan(file));
+
+          return next(null, processedValue);
+        });
+      }
+    ], cb);
+  }
+
+  static _processConfigFileValue(obj, baseFolder, fieldPath, done) {
+    if (Array.isArray(obj) || lodashIsObject(obj)) {
```
DennyGD: I think that `isObject` from lodash will return true for an array, although I haven't tested it

lib/ConfigFileProcessor.js
```diff
@@ -13,11 +13,16 @@
  * contents is a violation of applicable laws.
  */
 
+const fs = require('fs');
 const async = require('async');
-
-const { ConfigType } = require('./Constants');
+const path = require('path');
+const logger = require('./logger.js');
```
DennyGD: For better or worse, I've always envisioned that the `logger` module will be used through the `CLIManager`'s `log` method. 'Cause this is where the `logger` is 'initialized' depending on the options passed to the CLI by the user. And now I see that even there's additional logic in this `log` method. So I think that either some refactoring should occur or logging here should happen through the `log` method.

lib/ConfigFileProcessor.js
dokov: Makes sense, we wouldn't want this to print if there is an error, I will put it in an "if".

lib/ConfigFileProcessor.js
dokov: Not really, but will remove it :)

lib/ConfigFileProcessor.js
dokov: Good catch. Will change it to isPlainObject.

lib/ConfigFileProcessor.js
dokov: Hmmm, I copied this from Utils.js where the logger is used directly. In general it is a good practice for the logger to be a signleton, e.g. have a configured instance that is accessible from anywhere.

DennyGD: I'm not opposing the idea 😃 
The usage in Utils was a leftover meant to be changed. I don't think we should have logging in Utils functions, anyways.

PR #125, opened 2018-09-20
Cli 125 update service files structure

DennyGD:
Update service file structure to prepare for multiple environments in the future.
Example:
```
{
  "schemaVersion": "1.0.0",
  "configType": "service",
  "type": "flex-internal",
  "environments": {
     "dev": {
       "secret": "123",
       "sourcePath": "C:\\Work\\kinvey\\my-flex"
     }
  }
}
```

Environments could be a reference to a file, too. As this is still v2, the environments property is mandatory and it should contain 1 env.


test/ConfigManagementHelper.js
```diff
@@ -747,7 +750,7 @@ service.assertFlexServiceStatus = function assertFlexServiceStatus(id, expectedV
 
 service.assertFlexServiceStatusRetryable = function assertFlexServiceStatusRetryable(id, expectedVersion, expectedStatus, done) {
   async.retry(
-    { times: 6, interval: 20000 }, // 6 times every 20 sec
+    { times: 9, interval: 20000 }, // 6 times every 20 sec
```
dokov: Comment not changed :P

PR #128, opened 2018-10-03
Cli 121 org export

DennyGD:


lib/exporter/OrgFileExporter.js
```diff
@@ -0,0 +1,214 @@
...
+          services = data;
+          next();
+        });
+      },
+      (next) => {
+        if (isEmpty(services)) {
+          return setImmediate(next);
+        }
+
+        mkdirp(`${options.dir}/a`, next); // mkdirp expects filePath to create the directory
```
dokov: Shouldn't we put this inside the exportService function? It is best for the function itself to make sure the specified path exists, as opposed to making sure it is created in every place that you invoke the function.

DennyGD: Here, we create a service directory (only in case there are services to be exported) and we tell the exportService function where to save the exported service. The same function is used on the 'service export' command, so it expects a file path.

dokov: So, there is an exportService() function that accepts a filePath as an argument. There is a hidden requirement though - the directory for the given filePath must already be created, otherwise the function will fail. Is that correct?
What I am suggesting is making sure the necessary folders are created inside the exportService() function, so that there is no hidden requirements.

DennyGD: What you call a 'hidden requirement', I call 'expected behavior' :grin:
I think we have a slight communication gap. When I implemented the export commands I thought that the user must specify a path where all the directories already exist. I see now that when you implemented the env export, you thought that it is fine if a path with non-existing directories is specified...
Can I fix it here or should I create a ticket?

dokov: You should create a ticket. Doesn't mean you cannot fix the ticket in this PR. Up to you.

dokov:
Looks good, but integration tests are failing.

DennyGD:
The integration tests execute against staging and it is not the most stable environment, so it is kinda hard to decide whether there's a problem with the code or with the env itself. Most of the errors were mongo related: too many namespaces/collections. We've restarted the build and now it's all green: https://travis-ci.org/Kinvey/kinvey-cli/jobs/442168670

PR #136, opened 2018-10-26
Cli 151 support in app services

DennyGD:


test/integration-no-mock/config-management/app-config-create.js
```diff
@@ -14,6 +14,7 @@
  */
 
 const async = require('async');
+const cloneDeep = require('lodash.clonedeep');
```
dokov: I might be missing something, but this does not appear to be used.

DennyGD: Fixed

PR #144, opened 2018-11-19
CLI-162: Implemented support for RapidHealth services.

dokov:


lib/ServiceFileProcessor.js
```diff
@@ -248,9 +248,17 @@ class ServiceFileProcessor {
   static _buildRapidDataServiceObjectWoEnv(source, name, original = {}) {
     const result = getObjectByOmitting(source, ['configType', 'schemaVersion', 'type', 'environments']);
     result.name = name || source.name || original.name;
-    result.type = ConfigFiles.ConfigToBackendServiceType[source.type];
-    if (source.type === ConfigFiles.ServiceType.DATA_DIRECT) {
-      result.baseConfig = 'dataDirect';
+
+    if (source.baseConfig) {
```
DennyGD: How does this work? As far as I know `source` is the config file content, `name` is wanted entity name and `original` is the backend entity (if existent). So, `source` being the config file, it shouldn't contain a property called `baseConfig`, right? We should set `baseConfig` on result if `source.type` is dataDirect or the-health-thing... Or am I missing something here?

dokov: I decided to add support for baseConfig in general, as it seems to be a field that is used to specify some predefined base config. We know about two of those values and we handle them with special types. However, there might be other values in the future (or even now :) ) and I decided to add support for that.
Basically, it works like this:
ON EXPORT:
if we have baseConfig and it is one of the recognized values => transform it to the special type in the config file (e.g. { type: "rapid-health" }
if we have baseConfig, but the value is not recognized => put the baseConfig in the config file e.g. { type: "rest", baseConfig: "..." }

ON PUSH:
if we have a special type => transform it to { type: "rest", baseConfig: "..." }
if no special type, but baseConfig => put the baseConfig value in the payload

PR #148, opened 2018-11-27
Update readme

boevski:


README.md
```diff
@@ -614,6 +616,17 @@ Kinvey CLI is a subject to the following caveats:
 
 If problems persist, contact [Kinvey](http://support.kinvey.com).
 
+### Support for Service Environments
+
+Kinvey CLI version 4.1.0 introduces support for environments inside Flex services coinciding with the release of this feature on the backend. When upgrading from a previous Kinvey CLI version, have the following caveats in mind:
+
+- The `flex init` interactive command prompts for service environment selection in version 4.1.0 and later.
```
DennyGD: It won't prompt for a svc env if the service contains only one. Should we mention this?

README.md
```diff
@@ -614,6 +616,17 @@ Kinvey CLI is a subject to the following caveats:
...
+- Node.js projects configured with `flex init` with a Kinvey CLI version predating 4.1.0 do not contain a service environment ID. Therefore:
+    - We recommend rerunning the `flex init` command on these projects.
+    - Flex commands targeting the current service will continue to work as long as the service has a single environment.
+    - Flex commands targeting the current service will error out if you create additional environments inside the service.
```
DennyGD: and you don't specify a svc env or is this redundant?

README.md
boevski: Added.

README.md
boevski: Added.

PR #155, opened 2018-12-14
Feature clone appenv

timw255:


dokov:
Hey Tim, we are thinking about how this all fits with our plans for the future of CLI. We are working on the config management feature, which will also be able to do similar functionality. This is why this is taking us so long.

In the meantime, @DennyGD, can you review the code?

PR #157, opened 2019-01-03
Cli 124 org push

DennyGD:
* Add `org push <file> [--org <org-identifier>]` command
* Add tests
* Update README


lib/OrgFileProcessor.js
```diff
@@ -0,0 +1,240 @@
...
+  _createApps({ apps, orgId }, done) {
+    if (isNullOrUndefined(apps) || isEmpty(apps)) {
+      return setImmediate(done);
+    }
+
+    const appNames = Object.keys(apps);
+    async.eachSeries(
+      appNames,
+      (currName, next) => {
+        this.cliManager.log(LogLevel.INFO, `Creating app: ${currName}`);
```
dokov: Similar INFO message seems to be missing in the _updateApps method.

dokov: There are some failing integration tests, though.

dokov:
There are some failing integration tests, though.

lib/OrgFileProcessor.js
DennyGD: Yeah, this one turned out to be redundant, actually - the same info was being logged twice.

PR #165, opened 2019-04-09
Add default node version info.

boevski:


README.md
```diff
@@ -247,7 +247,7 @@ Kinvey CLI is distributed as an NPM package. After you install NPM, run the foll
         
     * `--runtime <node6|node8|node10>`
     
-        Specifies major Node.js version to run the project on. The minor and patch versions will vary depending on the [latest Flex Runtime updates](#flex-runtime-version-selection).
+        Specifies major Node.js version to run the project on. The minor and patch versions will vary depending on the [latest Flex Runtime updates](#flex-runtime-version-selection). Defaults to `node10`.
```
DennyGD: Have you considered using a link to devcenter only, without specifying a version here?
This way we would have a smaller amount of places where something must be changed. Also, it is kinda odd to update documentation for a client tool because of a server change. Moreover, if this PR gets merged, the npm documentation (https://www.npmjs.com/package/kinvey-cli) won't change until the CLI is published again. And it doesn't make much sense to get it published because of a README change.
What do you think about these things?

boevski: You know I'm a sucker for optimizing documentartion but in this case I feel it's very important to include the informalton here as well as in DevCenter because it's a stepping stone for choosing a Node version to develop against. Another reason why I vote for this approach is that the info is unlikely to change soon - Node does not release major versions so often, neither do we switch to one with similar frequency.

About NPM - yes, it does not make sence to release a new version because the change is not in the client, but it's an important change neverthemess. What do you suggest?

DennyGD: Yes, I agree that these type of changes don't come up often.
An alternative approach I think is to link to the devcenter info.
I suppose that this paragraph will be updated with the node.js version (https://devcenter.kinvey.com/rest/guides/flexservice-runtime#environment), so instead of having:

> Defaults to `node10`.

we could have some text pointing to the aforementioned link.
In this case, the info is present in the CLI's README file but it's not hard-coded there. Personally, I don't see any drawbacks. ❓

boevski: Thought of that, but it affects the user experience too much. Imagine having to click and go to a separate documentation set just to see those three words. The link text will probably be longer than that. What's more, we can't link directly to these words--we can only link to the section, which means the reader will have to read through it to find the version information.

DennyGD: I guess that if we decide we can link to the exact version...
What I don't like is that we can actually show wrong info here for some period of time. When we change the runtime version and don't publish the CLI, the users will see wrong/outdated information without knowing it. It's just that I find correctness more important than convenience.

Anyways, @dokov it's your call.

dokov: Maybe doing something in the lines of: Defaults to the Kinvey-recommended node version ('node10' as of this CLI release, read here[Denny's link] for latest info).

What do you think?

DennyGD: Sounds to me like an acceptable middle ground.

dokov: @boevski Agreed?

boevski: A bit combersome, but an acceptable compromise.

PR #167, opened 2019-04-15
Meta 327 sites linked to apps

DennyGD:
`website create` - require app or org
`website list` - show only ID + name
`website show` - show app or org

Note: Should be merged and published after the backend is deployed.

### DIFF
--- a/__snapshots__/website-create.test.js
+++ b/__snapshots__/website-create.test.js
@@ -1,3 +1,21 @@
+exports['website create using active profile with a name and app id should succeed and output JSON 1'] = `
+[debug] Checking for package updates
+[debug] Using profile 'activeProfile'
+[debug] Using application: 885f5d307afd4168bebca1a64f815c1e
+[debug] Request:  GET http://localhost:3234/v3/apps/885f5d307afd4168bebca1a64f815c1e
+[debug] Response: GET http://localhost:3234/v3/apps/885f5d307afd4168bebca1a64f815c1e 200
+[debug] Request:  POST http://localhost:3234/v3/sites
+[debug] Response: POST http://localhost:3234/v3/sites 201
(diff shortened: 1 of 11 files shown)


__snapshots__/website-create.test.js
```diff
@@ -74,6 +93,13 @@ Created website: 9caf90c31c4449f195a1a40acc979cf0
 
 `
 
+exports['website create using active profile with name, app name, historyApiRouting and errorPage should fail 1'] = `
```
dokov: Is this CLI logic? If yes, then it should be moved to the server. If not, we do not need a test for it, I think.

DennyGD: It's both server-side and client-side logic.
When you type `website create` the CLI creates a website as well as a site environment and this is isn't something known by the user. Therefore, it won't be nice if site creation succeeds but siteEnv creation fails. So, there's validation in the CLI for `historyApiRouting` and `errorPage` (which are siteEnv fields).

dokov: What I meant is that here you are testing that if you supply historyApiRouting=true and non-empty error page the command will fail. As far as I understand, this is not CLI logic (should not be). The CLI only supplies arguments to the server. This should be server logic and should be tested in the server.

You should have a test case where creation fails in order to ensure proper CLI error handling, but not have test cases for each case where you know the server will fail.

DennyGD: This is server-side logic.
But there is also validation in the CLI for the same thing - if `historyApiRouting` is `true`, then `errorPage` cannot be supplied. Why? Because users don't know/see that we actually have environments in the websites. From a user perspective `website create` is an atomic operation and only one entity is created. In reality, we create a website and then an environment. If env creation fails, things get a bit weird. To minimize the risk of env creation failing, we have client-side validation for routing + errorPage. Hence, the test in question.

dokov: OK, understood.

PR #172, opened 2019-08-07
Deprecate flex job.

boevski:


README.md
```diff
@@ -276,11 +276,11 @@ Kinvey CLI is distributed as an NPM package. After you install NPM, run the foll
     
 * `flex job [id]`
 
-   Shows the job status of a deploy/recycle command. If you don't specify an `id`, the command returns the status of the most recent `flex deploy` or `flex recycle` command.
+   _Deprecated_ Shows the job status of a deploy/recycle command. If you don't specify an `id`, the command returns the status of the most recent `flex deploy` or `flex recycle` command. _Deprecation note: For backward compatibility, the command always returns COMPLETE when executed without an `id`. Use `flex status` to track the progress of deploy or recycle jobs._
```
georgiwe: I think we shouldn't call them "jobs" (at the very end) since they are not jobs any more. And we should probably make this clearer, that these operations are not jobs. Deploy is a long-running process, which can be tracked through `flex status` and recycle affects the current status of the service, which is again tracked through `flex status`.

boevski: I understand your point but I do think implementation changes your perspective a bit. In any case, jobs or no jobs, we are deprecating `flex jobs`, which remains the only place where we refer to these as jobs.

I'm still changing the last occurrene of jobs though.

README.md
```diff
@@ -276,11 +276,11 @@ Kinvey CLI is distributed as an NPM package. After you install NPM, run the foll
     
 * `flex job [id]`
 
-   Shows the job status of a deploy/recycle command. If you don't specify an `id`, the command returns the status of the most recent `flex deploy` or `flex recycle` command.
+   _Deprecated_ Shows the job status of a deploy/recycle command. If you don't specify an `id`, the command returns the status of the most recent `flex deploy` or `flex recycle` command. _Deprecation note: For backward compatibility, the command always returns COMPLETE when executed without an `id`. Use `flex status` to track the progress of deploy or recycle commands._
```
DennyGD: > command always returns COMPLETE when executed without an `id`

maybe, the backend does always return `COMPLETE` but the CLI still checks if there's a previously saved job ID and if not it outputs the help menu + `This project is not configured. Use 'kinvey flex init' to get started. Alternatively, use positional arguments: id.`

georgiwe: The deprecation method is the id. It is the value of the _id field, returned by the server.

boevski: The rewording is out.

PR #180, opened 2019-11-07
META-449: Update readme

tsvetomir-nedyalkov:


README.md
```diff
@@ -85,6 +85,7 @@ Kinvey CLI is distributed as an NPM package. After you install NPM, run the foll
 * `org apply`
 
     Applies org configuration file to the specified org or to the active one.
+    Also applies organization services and organization apps configuration files if they are in the respective `services` and `applications` subfolders.
```
DennyGD: An org config file contains org-level services and apps, so applying an org file means modifying these services and apps. It is what the org config file is all about. 
I think that if the org config file needs clarifications they should be in the `Configuration management` section. The `Commands` section is just for command-specific info but you should know what is config management and this is the place to find how to apply the org config file for example. At least, this is the way I see it.

Edit: The same comment applies to the other modifications in the `Commands` section.

README.md
```diff
@@ -1149,13 +1244,23 @@ The following service template can be used and modified as needed:
...
  
-`environments.[envName].host` URI pointing to the service server's location. Required when `type` is 'flex-external'.
+`environments.[envName].host` URI pointing to the service server's location. Required for all service types except 'flex-internal' and 'rapid-health'.
 
 `environments.[envName].sourcePath` Path to source code - relative or absolute. Optional. Applicable when `type` is 'flex-internal'.
 
-`environments.[envName].environmentVariables` Environment variables. Object. Optional.
+`environments.[envName].runtime` The nodejs version of the runtime. Optional. The value must be one of [node6, node8, node10, node12]. Applicable when `type` is 'flex-internal'.
+
+`environments.[envName].authentication` Authentication type and credentials. Object. Optional. Applicable for RapidData services
```
DennyGD: I think that instead of RapidData we should specify the types as they are in the CLI (e.g. `rapid-health`). Otherwise, it's a bit confusing.

README.md
```diff
@@ -1149,13 +1244,23 @@ The following service template can be used and modified as needed:
...
+
+`environments.[envName].connectionOptions` Connection options. Object. Optional. Applicable for RapidData services
```
DennyGD: I think that instead of RapidData we should specify the types as they are in the CLI (e.g. rapid-health). Otherwise, it's a bit confusing.

README.md
```diff
@@ -1149,13 +1244,23 @@ The following service template can be used and modified as needed:
...
+
+`environments.[envName].mapping` Contains the source object, the source fields mapping and the supported operations. Object. Optional. Applicable for RapidData services
```
DennyGD: I think that instead of RapidData we should specify the types as they are in the CLI (e.g. rapid-health). Otherwise, it's a bit confusing.

README.md
DennyGD: Given that this is an object, should we list the properties?

PR #183, opened 2020-06-09
KDEV-680: Adjust to licensing changes

DennyGD:
* Change commands that previously accepted `--app <app>` or `--org <org>` to accept only `--org <org>`. If the `org` option is not specified, then the CLI checks if an active org has been set (`kinvey org use`). If an active org is not set, then an error is displayed.
Commands from the `flex` namespace first check if an org is specified in `.kinvey` in current directory before looking for the active org.
* Modified commands: `flex init`, `flex create`, `flex list`, `service create`, `website create`, `app create`.
* Adjusted config management tests.


DennyGD:
I'll open a PR to branch 'next-major-version', actually.

PR #192, opened 2020-09-04
KDEV-539: Add redirect page after cli login

ivanovit:
Add html page to redirect after login. All content is embedded in a single page(including the svg) because we want to send the page and close the server. If we have more files(svgs at separate files) we have to determine when all of the content is served and then to close the server.


bundyo (approved):

.vscode/launch.json
```diff
@@ -0,0 +1,22 @@
...
+            ],
+            "program": "${workspaceFolder}/bin/kinvey",
+            "console": "integratedTerminal",
+
+            // example commands
+            "args": ["init"],
+            // "args": [ "profile", "login"]
+        }
+    ]
+}
```
DennyGD, line 22: I'm reluctant to adding IDE-specific settings to source control.

DennyGD (changes requested):
It would be nice to have a Jira item ID for the PR and the commits.

.vscode/launch.json
ivanovit: Why do you see it as a problem?

DennyGD: Well, because it is IDE-specific, it could as well be dev-specific. It is not something that you **need** to run the project or build it, so no need to add it to source control. And even though I'm personally not using vscode I suspect it might be annoying for someone that uses it and wants to have a different configuration.

ivanovit: I agree with these arguments. But from a practical point of view, the debug/launch config should be common for the developers since it doesn't contain any personalizable options. Sometimes there are specific options in order to make it works - in this case, each developer will invest time for configuration. I believe that having the debug/launch configuration in source control makes development easier.

DennyGD: I suppose that the launch configuration has its own development path (hence the `version` in it) and I think that the configuration might differ from one developer to the other - e.g. one can choose from a variety of commands to put in `args`. 
Anyways, I feel that it is more important that we want to add to version control something that is not directly related to the project, nor required. It is a piece of code that will work only for a specific IDE. This doesn't feel like the best path.
What you're saying for making development easier though - I think that we could add some development info to the README file or to a different one. It might be beneficial as this is in the end a public repo and this lack of info is an omission on our part. Having said that I think that it might be better to include this improvement in a different PR as it is not in the scope of this one and to get this one moving forward.

DennyGD (changes requested):
The login page is far better this way, far better 😄 
There's a 'Callback already called' error, though that we should fix. It also prevents us from saving the profile (the connection settings).

DennyGD (approved):
I've removed the vscode folder as we need to move this PR forward. Adding some info to ease the development is always welcome though.

PR #196, opened 2020-10-26
Fix NotFound error with 12 character long Env ID

mrodrig:
This converts the allowed lengths to a list and checks that the provided environment ID's length matches one of the allowed lengths.

Fixes https://github.com/Kinvey/kinvey-cli/issues/195

### DIFF
--- a/lib/Utils.js
+++ b/lib/Utils.js
@@ -349,8 +349,8 @@ Utils.isEnvID = function isEnvID(value) {
   }
 
   // kid_SklZwh7dN
-  const lengthEnvId = 13;
-  const isId = value.length === lengthEnvId && value.startsWith('kid_');
+  const validEnvIdLengths = [12, 13];
+  const isId = validEnvIdLengths.includes(value.length) && value.startsWith('kid_');
   return isId;
 };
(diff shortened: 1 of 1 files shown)


mrodrig:
The integration test check seems to have failed for the same reason that the integration test check against `master` is failing:
  0 passing (5ms)
  1 failing
  1) Config management
       "before all" hook:
     Error: Env variables KINVEY_CLI_EMAIL(undefined) and KINVEY_CLI_PASSWORD must be set to execute the tests.
      at Context.<anonymous> (test/integration-no-mock/config-management/config-management.test.js:40:13)
      at processImmediate (internal/timers.js:456:21)

mrodrig:
Just updated this to remove the length check since it was mentioned in the support ticket I have open that the environment ID could also be 14 characters long. I haven't run into that, but it seems like the `kid_` prefix is the most reliable indicator that it's an environment ID.

DennyGD (approved):
The config management tests have failed because of our travis setup and this being a fork. Anyways, I've run them locally and they are good.
Thank you for your contribution!

PR #199, opened 2021-02-09
KDEV-1309: Modify hooks sequentially

DennyGD:


georgiwe (approved):
LGTM, though we could do updates on collections concurrently and only do sequential hook updates.

DennyGD:
> 
> 
> LGTM, though we could do updates on collections concurrently and only do sequential hook updates.

Yes, the collections are still processed in parallel: https://github.com/Kinvey/kinvey-cli/blob/5b8c441250b0d4ed3c959bd54e0dfc93107ff9ac/lib/EnvFileProcessor.js#L187

PR #200, opened 2021-03-04
KDEV-1510: Add more info for iOS push settings

DennyGD:


README.md
```diff
@@ -858,6 +858,8 @@ Before an environment config file gets applied a check will be executed to verif
 
 When the file is being applied to an existent environment then existent entities could be updated and new ones can be created. The CLI won't delete any entities even if they are not included in the file. Hence, if you want to create new endpoints but don't want to modify any existent collections for example, then you can exclude them from the configuration to have the file being applied faster.
 
+Please note that a config file generated by exporting an environment with iOS push settings configured, should be modified slightly before usage. Downloading an iOS push certificate previously uploaded to Kinvey is not possible. This is why the generated config file contains only its name (`push.ios.certificateFilename`) but not the certificate itself. To apply the exported config file, `push.ios.certificateFilename` must be replaced with `push.ios.certificateFilePath` (a path to the actual certificate). Another alternative is removing the `push.ios` field but this means that iOS push settings won't be applied or will be removed if the config file is applied to an environment with configured iOS push settings.
```
ggarabedian, line 861: I think we can go for something shorter here:

> :warning: Config files generated by the export command contain only the **name** of the iOS push notification certificate, and not the actual certificate. To apply the exported config file and have valid push settings, you need to replace `push.ios.certificateFilename` with `push.ios.certificateFilePath` - the path to the actual certificate on your machine. Another alternative is removing the `push.ios` field - however, this means that the iOS push settings won't be applied or will be removed if the config file is applied to an environment with configured iOS push settings.

ggarabedian (dismissed):

README.md
ggarabedian:
```suggestion
⚠️ Config files generated by the export command contain only the name of the iOS push notification certificate, and not the actual certificate. To apply the exported config file and have valid push settings, you need to replace push.ios.certificateFilename with push.ios.certificateFilePath - the path to the actual certificate on your machine. Another alternative is removing the push.ios field - however, this means that the iOS push settings won't be applied or will be removed if the config file is applied to an environment with configured iOS push settings.
```

ggarabedian (approved):

PR #201, opened 2021-04-02
KDEV-1570: Support for login hooks

alexanderfilipov:
* Allow new login hooks only for `user` collection.
* Updated unit and integration tests

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -671,7 +671,7 @@ Constants.CollectionHook = {
 };
 
 Constants.AllowedHooksUserCollection = Object.keys(Constants.CollectionHook);
-Constants.AllowedHooksCustomCollection = Constants.AllowedHooksUserCollection.filter(s => s.indexOf('Login') < 0);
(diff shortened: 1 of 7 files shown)


lib/Constants.js
```diff
@@ -660,14 +660,18 @@ Constants.CollectionHook = {
...
   onPostFetch: 'post-fetch',
-  onPostDelete: 'post-delete'
+  onPostDelete: 'post-delete',
+  onLoginSuccess: 'login-success',
+  onLoginFailure: 'login-failure'
 };
 
-Constants.AllowedCollectionHooks = Object.keys(Constants.CollectionHook);
+Constants.AllowedHooksUserCollection = Object.keys(Constants.CollectionHook);
+Constants.AllowedHooksCustomCollection = Constants.AllowedHooksUserCollection.filter(s => s.indexOf('Login') < 0);
```
DennyGD, line 674: I think that `s.includes` could be used.

DennyGD (commented):
Maybe, the tests should be reviewed. They usually are not very stable (because of stgus1) but I see a ServiceInUse error that shouldn't happen I think. 
Oh, and just as a note here - we shouldn't merge until the backend is released.

PR #203, opened 2021-11-04
KDEV-2048: Add password reset URL support for appenvs

DennyGD:
* Export the `passwordReset` field of the app environments and allow its presence in the env config file.


alexanderfilipov (approved):

alexanderfilipov:
How about some test?

